PR #81, opened 2020-08-15
fix: Support for useQueryState in dynamic routes

MRobertEvers:
Dynamic routes did not work because the `router.pathname` field includes the dynamic route name, rather than value.

E.g.
In a `/pages` folder of
```
/pages
  /views
    [view].tsx
```

```
>>> router.pathname
/views/[view]
```

Instead, use `router.asPath` and remove the fragment and query.

```
>>> router.asPath.split(/\?|#/, 1)
/view/my-great-view
```


MRobertEvers:
It may be better to use a built-in url parser, rather than a cute Regex. Unfortunately, there is nothing that parses just the pathname + query + fragment of a url. You could put a fake host in front of it and use the URL class.

Something like 

```
let url = new URL("http://localhost" + router.asPath);
```

Then `url.pathname` would return what we want, but I'm not sure. It looks suspicious doing that.

franky47:
Thanks for catching that, I'd say the Regex looks cleaner, it's not complex so I don't think it could cause performance or security issues.

MRobertEvers:
I found a problem the original implementation for dynamic path support.

```
      const [asPath] = router.asPath.split(/\?|#/, 1)
      updateUrl?.call(router, {
        pathname: asPath,
        hash: window.location.hash,
        search: query.toString()
      })
```

The above caused SSR each time; I didn't catch this originally. As per the nextjs docs, you have to use the `as` parameter. See 
74f2116

franky47:
Awesome, thanks ! I'm going to add unit tests to have meaningful CI and then we'll be good to go for a merge.

PR #288, opened 2022-02-28
Collect and merge multiple query updates into one

doxylee:
Hello, thank you for this great hook you've made! It was the best among other similar ones.
I was going to use it for my projects, but I needed to change some things for my requirements. I thought it would be nice to share what I've done.

So this problem with calling router.push at the same time causing overwrites was bugging me many times, so I created a solution to fix it. It removes the need for awating between query updates.

Basically what I'm doing is, "Collect query updates in a render, then merge and apply them at once at the start of the next render." My implementation uses Context API and `DeferredRouterProvider` must be an ancestor.

This is what happens in more detail:

1. `DeferredRouterProvider` is at the top of the component tree and provides `DeferredRouter` via the context API.
2. `useQueryState` gets `DeferredRouter` with `useDeferredRouter` hook and returns `setState`.
3. `setState` is called, and it calls `DeferredRouter.change` internally.
4. `DeferredRouter.change` causes `DeferredRouterProvider` to be rerendered after this render, and add the query update payload to a queue.
5. More calls on `DeferredRouter.change` follows, and query updates are all queued inside `DeferredRouter`
6. At the next render, `DeferredRouterProvider` is rendered foremost because it's the parent. And it calls `DeferredRouter.flush` to apply all the queued query updates.
7. The queued query updates are merged and applied with `Router.push` or `Router.replace`. It stops the current render phase, and starts a new render with the new query strings.

There are still things that needs to be done such as,
- Handle transitionOptions when query updates are merged.
- Optimizations
- Tests?

This is my first time contributing to open source and I'll try to give more info if you ask.

Thank you again for your great hook, and I hope to hear your feedback!


src/useQueryStates.ts
```diff
@@ -56,6 +57,7 @@ export function useQueryStates<KeyMap extends UseQueryStatesKeysMap>(
   { history = 'replace' }: Partial<UseQueryStatesOptions> = {}
 ): UseQueryStatesReturn<KeyMap> {
   const router = useRouter()
+  const deferredRouter = useDeferredRouter()
```
doxylee, line 60: I'm having a problem here.. I'm not familiar with how build process works on your environment.

image

doxylee:
@franky47  Thank you for your feedback! But the situation is quite complicated 😭

### Problem
By duplication of code, I believe you're referring to the part about `directUpdater` and `deferredUpdater` on `useQueryState(s)`, and `DeferredRouter.flush`.


There's some complication here.. Inside `directUpdater` (your original implementation), these things happen:
1. If `stateUpdater` is a function, get old values and apply `stateUpdater` to get new values.
2. Change the url by creating a new query string with `URLSearchParams` and call `router.push` or `router.replace`

But in `deferredUpdater`, the above steps are not happening inside `deferredUpdater`, but they are deferred and happens in `DeferredRouter.flush`. What actually happens in `deferredUpdater` is:

1. If `stateUpdater` is a function, wrap it with parse / serialize logic to create a `newStateUpdater` that handles states as strings. (This was needed because I thought `DeferredRouter` shouldn't know about parsing & serialization and it should be encapsulated inside the function that is handed to `DeferredRouter`.)
2. Queue the update payload to `DeferredRouter`.

Also, because of this, `DeferredRouter.push` is different from Next.js `Router` that it can also take functions as argument. This prevents swapping between Next.js `Router` and `DeferredRouter` like you suggested.


### Solution?
Regarding this problem, the best I can come up is allowing Next.js `Router` to also take a function as an argument. Actual implementation would be wrapping `useRouter` to create a custom hook `useNextRouter`(or some other name) which returns a augmented router that has push method which can take function as an argument. I think `directUpdater`'s logic should go inside `useNextRouter`, and deferredUpdater's logic should be the default inside `useQueryState(s)`. But doing this might mess up some of your `useMemo` optimizations..

I hope to hear what you think!

doxylee:
@franky47 I wasn't hearing from you for a long time, and I was wondering if it would be okay to implement what I suggested above. Waiting for your response, thanks.

franky47:
Hey @doxylee, sorry I don't have a lot of free time to work on this at the moment. I'll try and review your suggestion and come back to you.

If something can be done that adds functionality that does not increase bundle size for existing users (unless opt-in), then I could give it a closer look.

franky47:
That's great, thanks for mentioning the inspiration from this package.

I'll close this PR here then.

PR #314, opened 2022-12-13
fix: empty arrays include NaN after navigation

dopry:
I was implementing a filter that is an array of numbers, but was ended up with ['NaN'] in the array which was causing problems for code that expected a numeric id.  This seems to fix that issue.

### DIFF
--- a/src/defs.ts
+++ b/src/defs.ts
@@ -104,8 +104,11 @@ export const queryTypes: QueryTypeMap = {
     }
   },
   integer: {
-    parse: v => parseInt(v),
-    serialize: v => Math.round(v).toFixed(),
+    parse: v => parseInt(v) || null,
+    serialize: v => { 
+      if (isNaN(v)) return null;
+      return Math.round(v).toFixed();
+    },
     withDefault(defaultValue) {
       return {
         ...this,
@@ -114,8 +117,11 @@ export const queryTypes: QueryTypeMap = {
     }
   },
   float: {
-    parse: v => parseFloat(v),
-    serialize: v => v.toString(),
+    parse: v => parseFloat(v) || null,
+    serialize: v => { 
+      if (isNaN(v)) { return null }
+      return v.toString()
+    }
     withDefault(defaultValue) {
       return {
         ...this,


src/defs.ts
```diff
@@ -104,8 +104,11 @@ export const queryTypes: QueryTypeMap = {
     }
   },
   integer: {
-    parse: v => parseInt(v),
-    serialize: v => Math.round(v).toFixed(),
+    parse: v => parseInt(v) || null,
+    serialize: v => { 
+      if (isNaN(v)) return null;
```
franky47, line 109: Could you surround the check with curly braces please? To be consistant with the one below.

franky47 (changes requested):
Thanks! Some changes are required to handle zeros properly though.

src/defs.ts
```diff
@@ -114,8 +117,11 @@ export const queryTypes: QueryTypeMap = {
     }
   },
   float: {
-    parse: v => parseFloat(v),
-    serialize: v => v.toString(),
+    parse: v => parseFloat(v) || null,
+    serialize: v => { 
+      if (isNaN(v)) { return null }
+      return v.toString()
+    }
```
franky47, line 124: There's a coma missing here after the function body.

dopry: That's what I get for using the GH edit function to submit a quick PR.  I'm about to be out for the holidays, so I may not make it around to updating this before the new year.

franky47:
**note** A fix for arrays returning `[NaN]` has been merged in #330, the root problem had to do with letting empty strings through to parseInt / parseFloat, yielding NaNs.

I'd still like to keep your fix for not returning NaNs when parsing ints or floats, though I'll make some changes to conform to types. Eg: if you `setState(NaN)`, the serialised value in the querystring will be "NaN", as there's no way of bailing out of serialising.

franky47:
Rebased in #342.

dopry:
@franky47  hey, just wanted to say thank you for putting a bow on this. Sorry I couldn't see it through at the time.

Issue #327, opened 2023-05-12
Support for Next App Router (version 13.4)


andreisocaciu:
Hi, 

I use this library in my own projects and I was curious if there are any plans to add support for the new `app` directory that is now stable in [Next 13.4](https://nextjs.org/blog/next-13-4).

The main issue is that the `useRouter` from `'next/router'` is no longer supported in the `app` directory and it makes using this library impossible. It is being replaced by the `useRouter` from `'next/navigation'`, together with `usePathname` and `useSearchParams`.

franky47:
Eventually, yes, though I suppose this will require a major version bump because of the breaking changes you mention, or some sort of dual export to target either the pages or app routers.

That being said, with Next.js becoming more and more of a server-first framework, I expect a lot of hurdles along the way.

If you'd like to open a PR, I will gladly review it.

PR #328, opened 2023-05-15
feat: Update to useRouter from next/navigation

andreisocaciu:
Implemented the same functionality as before, but using the new Next.js 13.4 hooks from `'next/navigation'`.

It only works in the `app` directory now, so this means it should be released under v2.0 because of this breaking change.

Closes: #327


franky47:
Would there be a way to import this implementation in parallel to the previous one, using different import paths?

Eg:
```ts
// Using pages router
import { useQueryState } from 'next-usequerystate' // kept for compatibility
import { useQueryState } from 'next-usequerystate/pages' // explicit "legacy" import

// Using app router
import { useQueryState } from 'next-usequerystate/app'
```

franky47:
@andreisocaciu I tried updating the deps in https://github.com/47ng/next-usequerystate/commit/c46254a7, but somehow it didn't end up in this PR. Could you try replicating those changes and I'll approve the CI runs.

.github/workflows/cd.yml
```diff
@@ -17,7 +17,7 @@ jobs:
       - uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579
       - uses: actions/setup-node@56899e050abffc08c2b3b61f3ec6a79a9dc3223d
```
franky47, line 18: Sorry about that, I realised that the `actions/setup-node` must also be bumped to v3:

```yml
- uses: actions/setup-node@v3
  with:
    node-version: lts/*
```

franky47 (changes requested):

MaxLeiter:
FWIW `next/navigation` works in both pages and app. The types are different depending on if a `pages` folder is present -- in `pages` the new hooks can return `null` when `router.isReady` would be false.

simonexmachina:
Hello, I'm keen to use this in my project. Wondering if this will be merged soon, or if there's another fork that I can use?

luisanton-io:
This would be very helpful, hopefully merged soon...?

iloveitaly:
Same here!

franky47 (approved):

franky47:
I've merged these changes into `next` (somehow the PR got stuck in conflict limbo..), with some changes to make it possible to use both the app and pages router, using different imports:
```ts
import {} from 'next-usequerystate' // uses the pages router by default for compatibility
import {} from 'next-usequerystate/pages' // explicit opt-in to pages router
import {} from 'next-usequerystate/app' // explicit opt-in to app router
```

I published `1.8.0-beta.1` to test it on my website that uses the app router, feel free to test that version and report back!

tryanmac:
Works like a dream for me thanks all contributors!

PR #330, opened 2023-05-31
prevent empty string while setting an empty array as value

prasannamestha:
Fixes this issue: https://github.com/47ng/next-usequerystate/issues/329

### DIFF
--- a/src/defs.ts
+++ b/src/defs.ts
@@ -197,7 +197,9 @@ export const queryTypes: QueryTypeMap = {
           .split(separator)
           .map(item => decodeURIComponent(item))
           .map(itemSerializers.parse)
-          .filter(value => !value) as ItemType[]
+          .filter(
+            value => !value !== null && value !== undefined && value !== ''
+          ) as ItemType[]
       },
       serialize: values =>
         values
--- a/src/defs.ts
+++ b/src/defs.ts
@@ -195,10 +195,9 @@ export const queryTypes: QueryTypeMap = {
         type ItemType = NonNullable<ReturnType<typeof itemSerializers.parse>>
         return query
           .split(separator)
-          .filter(item => item !== '') // ignore empty items
           .map(item => decodeURIComponent(item))
           .map(itemSerializers.parse)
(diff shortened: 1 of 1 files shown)


franky47 (changes requested):
**nitpick**: This may have unintended consequences, like dropping zeros from arrays of numbers.

**suggestion**: Performing a test on the query value to check for an explicit empty string would fix the issue while maintaining existing "falsy but not nullish" operations.

src/defs.ts
```diff
@@ -197,7 +197,9 @@ export const queryTypes: QueryTypeMap = {
           .split(separator)
           .map(item => decodeURIComponent(item))
           .map(itemSerializers.parse)
-          .filter(value => !value) as ItemType[]
+          .filter(
+            value => !value !== null && value !== undefined && value !== ''
```
franky47, line 201: **nitpick**: This won't do either, as arrays of multiple string items can include empty elements. The heart of the issue comes from calling `.split` on an empty query string (not individual elements, but the whole query), so this case needs to be taken care of even before the query string is splitted/mapped/filtered.

Also, watch out for that stray negation on the null test.

franky47 (changes requested):

src/defs.ts
prasannamestha: You are right.. Will try to take care of this

prasannamestha: My apologies for the stray negation, seems like a cat walked on my keyboard 😺 

Regarding the heart of the issue - isn't it caused because .split operation returns an array with empty string (`''.split(',')`)?

franky47: Yes indeed. When the `query` argument of the `parse` function is an empty string, an empty array should be returned directly, in order not to hit the `.split` call. Otherwise, the split/map/filter logic can perform its task.

prasannamestha: Gotcha 👍 Updated the code

franky47 (approved):
**praise**: LGTM, thanks!

prasannamestha:
@franky47 thanks for merging the PR. Could you please let me know when you plan to release this fix?

PR #374, opened 2023-10-24
feat: Configurable throttling

franky47:
Adds a `throttleMs` option to slow down the rate of updates to the URL.

Note: internal state (the first item returned by the hook) is not throttled, and behaves just like React.useState to guarantee UI reactivity. Only **query string updates** to the URL are throttled, to handle browsers rate-limiting the history API, and to slow down sending network requests to the server when using `shallow: false`.

As an example, to safely bind a high-frequency query state on Safari, which has higher rate limits:

```tsx
function Search() {
  const [query, setQuery] = useQueryState(
    'q',
    parseAsString
      .withOptions({ throttleMs: 350 })
      .withDefault('')
  )
  return (
    <input
      value={query}
      onChange={e => setQuery(e.target.value)} 
    />
  )
}
```

See discussion #373.

### DIFF
--- a/README.md
+++ b/README.md
@@ -272,6 +272,35 @@ const [state, setState] = useQueryState('foo', { scroll: true })
 setState('bar', { scroll: true })
 ```
 
+### Throttling URL updates
+
+Because of browsers rate-limiting the History API, internal updates to the
+URL are queued and throttled to a default of 50ms, which seems to satisfy
+most browsers even when sending high-frequency query updates, like binding
+to a text input or a slider.
+
+Safari's rate limits are much higher and would require a throttle of around 340ms.
+If you end up needing a longer time between updates, you can specify it in the
+options:
+
+```ts
+useQueryState('foo', {
+  // Send updates to the server maximum once every second
+  shallow: false,
+  throttleMs: 1000
+})
+
+// You can also pass the option on calls to setState:
+setState('bar', { throttleMs: 1000 })
+```
+
+> Note: the state returned by the hook is always updated instantly, to keep UI responsive.
+> Only changes to the URL, and server requests when using `shallow: false`, are throttled.
+
+If multiple hooks set different throttle values on the same event loop tick,
+the highest value will be used. Also, values lower than 50ms will be ignored,
+to avoid rate-limiting issues. [Read more](https://francoisbest.com/posts/2023/storing-react-state-in-the-url-with-nextjs#batching--throttling).
+
 ## Configuring parsers, default value & options
 
 You can use a builder pattern to facilitate specifying all of those things:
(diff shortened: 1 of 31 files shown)


ShahriarKh:
I tried with a text input and noticed two problems:
1. The page gets re-rendered on every keystroke.
2. The value gets out of sync but sometimes shows the correct text after that (only with `shallow: false`). Here's a video:

https://github.com/47ng/next-usequerystate/assets/31452340/0c7a5f43-9e04-4100-8647-7b68a22f1a4e

 In the video above, left input uses `throttleMs`, right input uses the code mentioned at #373. (both are `1000ms`)

franky47:
> 1. The page gets re-rendered on every keystroke.

This is expected. Internally, this behaves much like React.useState, which would re-render when the state changes.

If you wish to limit the level of re-rendering, using a `<Suspense>` boundary will limit where the client code runs, which is what gets re-rendered. There's no need to have the hook at the root of the page, it can be deeply nested and duplicated where it's useful (eg: writing in one component and spying on the query value in others).

For point 2, this is probably a cache issue with the RSC client cache, I'll have a look, thanks for the report!

franky47:
Ok so the second point is actually a race condition between local edits and the RSC payload coming in that triggers a URL change to match what was rendered on the server.

So you type "foo bar". The first throttled request to the server sees and renders "foo", re-updates the local URL to "foo" and " bar" is dropped.

Throttling actually makes things worse because all edits between two responses may be lost, and increasing throttling increases the chance of it happening.

So it looks like the best approach is indeed to use an uncontrolled component, set the default value to the state value, which will deal with page loads / mounts, and use the `throttleMs` option to control how frequently we tell the server about URL updates.

franky47:
After playing for a while with throttling options and how to properly schedule server updates, I don't think there's an ideal setting that can work anywhere anytime.

Taking care of the race-condition of a stale external navigation overwriting the hooks internal states was easy once the right cause was identified: the URL update queue is now merged with incoming external search params to sync internal states.

However, it looks like RSC streams don't get applied to the DOM if the URL that triggered them is different when they arrive (probably due to the client RSC cache being keyed by the URL). This caused the whole app to be unresponsive when using only the Next.js router for query updates, as the URL would only be updated when the appropriate RSC payload resolved, which may never happen if the network connection is poor (or absent) and/or the serverless function times out on the backend.

Instead, query updates now **always** update the query string locally first (using the Web History API), then notify the server with a `replaceState` on non-shallow updates. This keeps the URL in sync with the UI, and lets the server catch-up asynchronously and stream-in RSCs.

For high frequency non-shallow updates (like binding a query state to a text input or a range slider), a higher `thottleMs` is recommended. To keep server-streamed UI updates responsive, it should be set higher than the server round-trip latency (ie: the time it takes to send the request, let the server render RSCs and stream the response back).

This behaviour is showcased in the [throttling playground](https://next-usequerystate-n9g1wv9np-47ng.vercel.app/demos/throttling), where artificial delays can be configured. Using the browser devtools to throttle the network connection also helps get an idea of how the server responds to floods of requests.

PR #398, opened 2023-11-16
Switch debug logging to `console.log`

tordans:
* Change the debugging logging from `console.debug` to `console.log` so logging is visible without changing the log level to "verbose"
* Simplify readme wording in the debug section

### DIFF
--- a/README.md
+++ b/README.md
@@ -479,9 +479,7 @@ See issue #259 for more testing-related discussions.
 ## Debugging
 
 You can enable debug logs in the browser by setting the `debug` item in localStorage
-to `next-usequerystate`, and reload the page. 
-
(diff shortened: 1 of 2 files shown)


franky47:
To be fair it made sense to use `console.debug`in early versions of the library when debugging info was always printed out, but we could go with `console.log` now (and save two bytes along the way).

Would you like to change your PR to do that? 

Relevant code is in `packages/next-usequerystate/src/debug.ts`

tordans:
@franky47 I updated the PR. I can squash the commits, if you like. But that should also work with the Github Squash-Merge if that is enabled.

Anything else here?

franky47 (approved):
LGTM, I'll do the squash and merge to use the proper SemRel semantics. Thanks!

PR #401, opened 2023-11-16
bug_report template: Use emoji as checkboxes

tordans:
The checkboxes will show up as "tasks" on the issue summary "1 of 2 tasks done" wich is misleading since they are not tasks.

(https://github.com/ikatyang/emoji-cheat-sheet/blob/master/README.md)

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -24,10 +24,10 @@ Next.js information (obtained by running `next info`):
 
 Are you using:
 
-<!-- Remove the ✅ (is used) or the ✖️ (is not used): -->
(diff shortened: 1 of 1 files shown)


franky47:
I don't particularly mind the tasks thing, however the comment describing what to do here is confusing.

Maybe something like "Keep whichever is relevant (✅: used, ❌: not used)" ?

franky47 (changes requested):
I don't particularly mind the tasks thing, however the comment describing what to do here is confusing.

**suggestion**: Maybe something like "Keep whichever is relevant (✅: used, ❌: not used)" ?

franky47 (approved):

franky47:
LGTM, thanks!

PR #406, opened 2023-11-17
feat: Add support for transitions

### LINKED ISSUE #402: Getting pending state of the router
Hi, there,

Thank you for creating this amazing package.

Not sure if you have encountered the following use case: 

Client component setting the state
```ts
const [searchValue, setSearchValue] = useQueryState(
 'client',
  parseAsString.withOptions({ shallow: false }).withDefault(''),
);
```

Server component fetching data based on the query state

```ts
const clients = await api.getClients({searchParam: queryState});
```
I am trying to get the Router `isPending` state to display a loading message, while the client component is waiting for the new data to be passed down from the server.

At the moment I am using `useTransition` in my code to achieve this.
```ts
  const [isPending, startTransition] = useTransition();

  const setSearchParameters = useCallback(
    (name: string, value: string) => {
      startTransition(() => {
        router.replace(pathname + '?' + createQueryString(name, value));
      });
    },
    [createQueryString, pathname, router],
```
Not sure if this is a feature you would like to consider to implement.

franky47:
This lets non-shallow updates be notified of server rendering loading status, using an external `React.useTransition` hook.

Example usage:

```tsx
function ClientComponent() {
  // 1. Provide your own useTransition hook:
  const [isLoading, startTransition] = React.useTransition()
  const [query, setQuery] = useQueryState(
    'query',
    // 2. Pass the `startTransition` as an option:
    parseAsString().withOptions({ startTransition })
  )
  // 3. isLoading will be true while the server is re-rendering and streaming RSC payloads
  // when the query is updated via `setQuery`.
}
```

## Tasks

- [x] Investigate dangling `startTransition` functions in unmounted components (see below)
- [x] Improve type safety around the interaction of the `shallow` and `startTransition` options. Done in #410, thanks @Talent30 !
- [x] Add end-to-end tests
- [x] Add documentation

Things that need to be ironed out:

What happens if we were to call a `startTransition` on an unmounted component? Because the URL update is deferred, a query state update may cause the calling component to be unmounted, and its `startTransition` reference left dangling in the queue. ~This is an issue with state updates, is it the same for transitions, and how can we mitigate it?~ -> This is no longer an issue in React 18 apparently: https://github.com/reactwg/react-18/discussions/82

Closes #402.

### DIFF
--- a/packages/next-usequerystate/src/defs.ts
+++ b/packages/next-usequerystate/src/defs.ts
@@ -1,10 +1,18 @@
 import { useRouter } from 'next/navigation.js' // https://github.com/47ng/next-usequerystate/discussions/352
+import type { TransitionStartFunction } from 'react'
 
 export type Router = ReturnType<typeof useRouter>
 
 export type HistoryOptions = 'replace' | 'push'
 
-export type Options = {
+// prettier-ignore
+type StartTransition<T> = T extends false
+  ? TransitionStartFunction
+  : T extends true
+    ? never
+    : TransitionStartFunction
+
+export type Options<Shallow = unknown> = {
   /**
    * How the query update affects page history
    *
@@ -29,7 +37,7 @@ export type Options = {
    * Setting it to `false` will trigger a network request to the server with
    * the updated querystring.
    */
-  shallow?: boolean
+  shallow?: Extract<Shallow | boolean, boolean>
 
   /**
    * Maximum amount of time (ms) to wait between updates of the URL query string.
@@ -48,8 +56,10 @@ export type Options = {
    * `React.useTransition()` hook.
    *
    * Using this will set the `shallow` setting to `false` automatically.
+   * As a result, you can't set both `shallow: true` and `startTransition`
+   * in the same Options object.
    */
-  startTransition?: React.TransitionStartFunction
+  startTransition?: StartTransition<Shallow>
 }
 
 export type Nullable<T> = {
--- a/packages/next-usequerystate/src/defs.ts
+++ b/packages/next-usequerystate/src/defs.ts
@@ -41,6 +41,15 @@ export type Options = {
    * will not have any effect.
    */
   throttleMs?: number
+
+  /**
+   * Opt-in to observing Server Component loading states when doing
+   * non-shallow updates by passing a `startTransition` from the
+   * `React.useTransition()` hook.
+   *
+   * Using this will set the `shallow` setting to `false` automatically.
+   */
+  startTransition?: React.TransitionStartFunction
 }
 
 export type Nullable<T> = {
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@ useQueryState hook for Next.js - Like React.useState, but stored in the URL quer
 - ♊️ Related querystrings with [`useQueryStates`](#usequerystates)
 - 📡 [Shallow mode](#shallow) by default for URL query updates, opt-in to notify server components
 - 🗃 _**new:**_ [Server cache](#accessing-searchparams-in-server-components) for type-safe searchParams access in nested server components
+- ⌛️ **new:** Support for [`useTransition`](#transitions) to get loading states on server updates
 
 ## Installation
 
@@ -317,6 +318,36 @@ setState('bar', { throttleMs: 1000 })
 If multiple hooks set different throttle values on the same event loop tick,
 the highest value will be used. Also, values lower than 50ms will be ignored,
 to avoid rate-limiting issues. [Read more](https://francoisbest.com/posts/2023/storing-react-state-in-the-url-with-nextjs#batching--throttling).
+
+### Transitions
+
+When combined with `shallow: false`, you can use the `useTransition` hook to get
+loading states while the server is re-rendering server components with the
+updated URL.
+
+Pass in the `startTransition` function from `useTransition` to the options
+to enable this behaviour _(this will set `shallow: false` automatically for you)_:
+
+```tsx
+'use client'
+
+import React from 'react'
+import { useQueryState, parseAsString } from 'next-usequerystate'
+
+function ClientComponent({ data }) {
+  const [isLoading, startTransition] = React.useTransition()
+  const [query, setQuery] = useQueryState(
+    'query',
+    parseAsString().withOptions({ startTransition })
+  )
+
+  // Indicate loading state
+  if (isLoading) return <div>Loading...</div>
+
+  // Normal rendering with data
+  return <div>{/*...*/}</div>
+}
+```
 
 ## Configuring parsers, default value & options
 
--- a/README.md
+++ b/README.md
@@ -335,11 +335,15 @@ import React from 'react'
 import { useQueryState, parseAsString } from 'next-usequerystate'
 
 function ClientComponent({ data }) {
+  // 1. Provide your own useTransition hook:
   const [isLoading, startTransition] = React.useTransition()
   const [query, setQuery] = useQueryState(
     'query',
+    // 2. Pass the `startTransition` as an option:
     parseAsString().withOptions({ startTransition })
   )
+  // 3. `isLoading` will be true while the server is re-rendering
+  // and streaming RSC payloads, when the query is updated via `setQuery`.
 
   // Indicate loading state
   if (isLoading) return <div>Loading...</div>
--- a/packages/e2e/cypress/e2e/transitions.cy.js
+++ b/packages/e2e/cypress/e2e/transitions.cy.js
@@ -1,0 +1,20 @@
+/// <reference types="cypress" />
+
+it('transitions', () => {
+  cy.visit('/app/transitions')
+  cy.contains('#hydration-marker', 'hydrated').should('be.hidden')
+  cy.get('#server-rendered').should('have.text', '{}')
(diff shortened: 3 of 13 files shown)


packages/next-usequerystate/src/defs.ts
```diff
@@ -41,6 +41,15 @@ export type Options = {
...
   throttleMs?: number
+
+  /**
+   * Opt-in to observing Server Component loading states when doing
+   * non-shallow updates by passing a `startTransition` from the
+   * `React.useTransition()` hook.
+   *
+   * Using this will set the `shallow` setting to `false` automatically.
+   */
+  startTransition?: React.TransitionStartFunction
```
tordans, line 52: Assuming that his are the main types for the package, it would be beneficial to make this interaction between `startTransition` and `shallow` part of the types.

Something like this should be possible: https://www.typescriptlang.org/play?ts=5.2.2#code/C4TwDgpgBAEglgZ2AewE4gPJmHZA7BKAXigEM8QAoSiADzDWClEiix30JIG9KooAFohToA-AC5YwtJmy4CAbj5QEAY1TIANpolQARsi0RyS-sAEbgwTRACyCXXgCuAWz0RUSgL5QAZFAAKZQAfKG4VAVJtZAB3XWBUJwgFFWBSVGAAFVRyBDgOPEcIADcPKC8QsIiozVjdADMohGhQvBKPFKR0rJyCfPldACVjVWAAOmzc-vwAZTSMgDEnPFH5csoASiVKVU4mesMARkl2eS4whEjomPEEpIAaVO7JvoLxAMCN4gA+MJ8v0iEYakUYTXp5ApzbpLFYFCoAenh-Ey4AghGQ9SgYA0kAyICgAHIuhkXhD5ASyKhoHAVsgXGBSDg9DYxpREfxkajCaTpngoYtlqt8BTEFA8MgmIC8gBzPCkZnQFDMLkE5YAEwg9RpEDVBLGAQATABmA0GjbUXYEfaGA0nOScYgXeY9KZvD4BL5EX7cf5kIEjcY8yHOmFCvAIpEYADSFr2UAOyCNdoK524lxqsUkjU0zUexJdr3k70+Pz+UAB-pBgfBvP5wFDcLZkZjOzjCYALMmzo601dajd400IBG2DGgA

franky47: This was suggested by @Talent30 in https://github.com/47ng/next-usequerystate/issues/402#issuecomment-1815755845, thanks for doing the heavy lifting for types!

Talent30:
For unmounted components I don't think there is a elegant solution at the moment.

I can only think of using a ref to keep track of the component's mounting status.

Will React throw an error if we call startTransition in an unmounted component?

franky47:
> Will React throw an error if we call startTransition in an unmounted component?

It doesn't seem to. But then again neither does setting state on an unmounted component either, is this something that's legal in React now?

Talent30:
> > Will React throw an error if we call startTransition in an unmounted component?
> 
> No, it doesn't seem to. Neither does it throw an error if setting state on an unmounted component, is this something that's now legal in React? I must have missed the memo.
> 
> Here's my test bench:
> 
> ```tsx
> 'use client'
> 
> import { parseAsBoolean, useQueryState } from 'next-usequerystate'
> import React from 'react'
> 
> const parser = parseAsBoolean.withDefault(true)
> 
> export function Client() {
>   const [showBench, setShowBench] = useQueryState('showBench', parser)
>   return (
>     <>
>       {showBench && <TransitionBench />}
>       {!showBench && <button onClick={() => setShowBench(true)}>Reset</button>}
>     </>
>   )
> }
> 
> function TransitionBench() {
>   const [, startTransition] = React.useTransition()
>   const [, setShowBench] = useQueryState('showBench', parser)
>   const [state, setState] = React.useState('idle')
>   console.log('render TransitionBench', state)
> 
>   React.useEffect(() => {
>     console.log('Mounting TransitionBench')
>     return () => {
>       console.log('Unmounting TransitionBench')
>     }
>   }, [])
> 
>   return (
>     <>
>       <button
>         onClick={() => {
>           setTimeout(() => {
>             startTransition(() => {
>               console.log('In the transition body')
>               // This should fail, but it doesn't
>               setState('unmounted?')
>             })
>           }, 500)
>           setShowBench(false)
>         }}
>       >
>         Start Transition & Remove component
>       </button>
>       <p>{state}</p>
>     </>
>   )
> }
> ```

Very odd, in my past experience it should throw `Can't perform a React state update on an unmounted component`, is that because we put it under `startTransition` and somehow it handles it?

franky47:
@Talent30 no apparently it's now OK to do that: https://github.com/reactwg/react-18/discussions/82

In a memory management point of view, what we're doing here is safe: we're storing references of the `startTransition` function in a queue, that will get emptied when the URL updates.

When a component unmounts, the only reference left is the one in the queue. When the queue gets flushed, it won't do anything particular wrt/ transitions, but once the URL update function exits, the last `startTransition` reference gets cleared and it can get garbage-collected.

Talent30:
A draft doc.

# React Server Components: useTransition with useQueryState

React Server Components allow for server-side rendering of components in a way that seamlessly integrates with client-side interactions. To enhance the user experience during the data streaming, developers can utilise the `useTransition` hook to manage loading states.


## Integrating `useTransition` with Server Component Updates

To integrate `useTransition` with server components, you can track the loading status of server component by passing the `startTransition` function to an external state management hook, such as `useQueryState`. During server-side rendering, the `isLoading` state will reflect whether the server is streaming updated React Server Components (RSC) payloads to the client. The `setQuery` function is used to trigger a query state update which will cause a server re-rendering if required.

## Usage

The following steps illustrate how to use `useTransition` with server-rendered components:

### Step 1: Use the `useTransition` Hook

First, call the `useTransition` hook in your client component to obtain the `isLoading` state and `startTransition` function.

```jsx
const [isLoading, startTransition] = React.useTransition();
```

### Step 2: Pass `startTransition` to the External State Hook (`useQueryState`)

Then, add `useQueryState` hook and pass the `startTransition` function as an option. Please note when you pass `startTransition` as an option the `shallow` mode will be set to `off`.

```jsx
const [query, setQuery] = useQueryState(
  'query',
  parseAsString().withOptions({ startTransition })
);
```

### Step 3: Use `isLoading` to Indicate Loading Status

With `isLoading`, you can show loading indicators or placeholders in your component's UI to indicate that the server is processing the updates and streaming to the client.

```jsx
if (isLoading) {
  return <LoadingIndicator />;
}
```

### Full Component Example:

```jsx
import React from 'react';
import { useQueryState, parseAsString } from 'next-usequerystate ';

function ClientComponent() {
  const [isLoading, startTransition] = React.useTransition();
  const [query, setQuery] = useQueryState(
    'query',
    parseAsString().withOptions({ startTransition })
  );

  // Perform actions, event handlers, etc., that may call setQuery when necessary.

  if (isLoading) {
    return <LoadingIndicator />;
  }

  // Render your component with the obtained query data, etc.
}

export default ClientComponent;
```

franky47:
> Not sure if data streaming is the right word to use

@Talent30 Not really, I did a test where the page component would resolve after 500ms, and includes a server component under a Suspense boundary that resolves after 1000ms, and expected the page update to be streamed in first, then the child RSC to be streamed in last and resolve the transition loading state, but it all arrives in one go. 🤷‍♂️ 

Thanks for your doc draft, it's a bit verbose though, the feature is simple enough to directly show a code example (and an interactive demo once the docs site is up).

packages/next-usequerystate/src/defs.ts
franky47: While it is indeed possible to represent the Options object this way, it makes consuming it in a generic way impossible within the hooks, without resorting to `// @ts-expect-error`. When we merge options declared at the hook level with options passed as arguments of the state updater function, we just run a simple merge, and let the update queue decide how to merge the result with the queue options. But this excluding union doesn't have any valid intersection (`{ shallow?: boolean; startTransition?: ReactTransitionStartFunction}`, so TypeScript complains.

I wondered about allowing `startTransition` in shallow mode, but the loading state remains `false` in that case, unsurprisingly. I'll have to see if I can find a valid use-case for using the raw `startTransition` function (without a loading observer), to mark a shallow URL update (and associated scroll) as interruptible by other UI events.

In the mean time, I'll probably move on to merging this without the discriminated union type, and let the documentation explain the behaviour. What do you think @tordans @Talent30 ?

Talent30: If a minimum of type safety can be ensured that would be great otherwise it shouldn't block this PR from moving on.
```
(
  | { shallow?: true; startTransition?: never }
  | { shallow?: false; startTransition?: React.TransitionStartFunction }
)
```

Talent30:
> > Not sure if data streaming is the right word to use
> 
> @Talent30 Not really, I did a test where the page component would resolve after 500ms, and includes a server component under a Suspense boundary that resolves after 1000ms, and expected the page update to be streamed in first, then the child RSC to be streamed in last and resolve the transition loading state, but it all updates in one go after 1000ms. 🤷‍♂️
> 
> Thanks for your doc draft, it's a bit verbose though, the feature is simple enough to directly show a code example (and an interactive demo once the docs site is up).

I have shortened the doc

PR #407, opened 2023-11-17
Github: Add links to discussion to new issue screen

tordans:
This will expand the UI on the new issues screen to direct users to use the discussion for feature requests.

You can see how that looks for example here https://github.com/openstreetmap/iD/issues/new/choose ([Template](https://github.com/openstreetmap/iD/blob/develop/.github/ISSUE_TEMPLATE/config.yml))


franky47 (approved):
LGTM, thanks!

PR #411, opened 2023-11-23
chore: update dependencies

Talent30:
Update package dependencies, not sure why CI is not running.


franky47:
Thank you, but there is already the Depfu bot taking care of this on regular intervals.

PR #412, opened 2023-11-23
fix: 'Source on GitHub' urls

psdewar2:
### What happened?
'Source on GitHub' urls in each demo broke as a result of moving files to support the monorepo:
image

### What changed?
I updated the urls to files in the playground folder.

### Validation
Urls point to the correct files in the repo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,15 +21,15 @@ useQueryState hook for Next.js - Like React.useState, but stored in the URL quer
 ## Installation
 
 ```shell
-$ pnpm add next-usequerystate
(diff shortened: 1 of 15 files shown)


README.md
franky47, line 1: **suggestion**: To allow copy-pasting individual commands, maybe we should drop the leading `$ `.

franky47 (approved):
LGTM, thanks for catching this!

README.md
psdewar2: done!

franky47: Thanks!

PR #419, opened 2023-12-02
fix: Push

franky47:
History: push is currently broken on Next.js 14.0.3 in the app router without the WHS flag, and has been broken for a while in the pages router.

Adding tests to cover those cases, and looking for ways to solve it, though in the case of the app router it may depend on https://github.com/vercel/next.js/pull/58861 to be made available.

Getting a consistent fix in the pages router that doen't break dynamic routes or basePath support is tricky. Initially I thought it was pointless to go through the History API in the pages router, as it supports shallow routing, but always going through the router without extra care seems to break in some combinations of settings and situations.


franky47:
Still some issues with basePath on the pages router, where basePath is applied twice. Relevant Next.js issues:
- https://github.com/vercel/next.js/issues/47486
- https://github.com/vercel/next.js/issues/54770

PR #429, opened 2023-12-11
fix: Use a traditional useEffect for emitter registration

franky47:
useInsertionEffect was a hack to be notified from navigation events in the same event loop tick as the Next.js router, but it causes issues with tRPC's SSR and it looks like we could get by with a simple useEffect.

Testing this in CI to see if it's a viable fix, and there'll be internal docs and scheduling code to update if all goes green.


franky47:
Closing as it makes sync with navigation difficult. Original issue had to do with tRPC using an internal library not compatible with React 18.

Maybe we should mark a peer dep on `react@18`, or whatever canary version Next.js vendors.

PR #453, opened 2024-01-11
feat: Add literal parsers

timheerwagen:
Hello, while testing with the package I noticed that there is no support for as const string arrays.

Since I am one of those people who don't like enums, these are better for me to use.

I thought this parser might be interesting for everyone.

I have also extended the documentation and adapted outdated documentation for parseAsStringEnum.

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,7 +103,8 @@ import {
   parseAsIsoDateTime,
   parseAsArrayOf,
   parseAsJson,
-  parseAsStringEnum
+  parseAsStringEnum,
+  parseAsStringConst
 } from 'nuqs'
 
 useQueryState('tag') // defaults to string
@@ -127,6 +128,15 @@ const [direction, setDirection] = useQueryState(
   'direction',
   parseAsStringEnum<Direction>(Object.values(Direction)) // pass a list of allowed values
     .withDefault(Direction.up)
+)
+
+// readonly Array (string-based only)
+const colors = ['red', 'green', 'blue'] as const
+
+const [color, setColor] = useQueryState(
+  'color',
+  parseAsStringConst(colors) // pass a readonly list of allowed values
+    .withDefault('red')
 )
 ```
 
@@ -594,7 +604,7 @@ export function Server() {
 
 // client.tsx
 // prettier-ignore
-;'use client'
+'use client'
 
 import { useQueryStates } from 'nuqs'
 import { coordinatesParsers } from './searchParams'
--- a/README.md
+++ b/README.md
@@ -104,7 +104,8 @@ import {
   parseAsArrayOf,
   parseAsJson,
   parseAsStringEnum,
-  parseAsLiteral
+  parseAsStringLiteral,
+  parseAsNumberLiteral
 } from 'nuqs'
 
 useQueryState('tag') // defaults to string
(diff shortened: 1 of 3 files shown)


packages/nuqs/src/parsers.ts
```diff
@@ -230,6 +229,40 @@ export function parseAsStringEnum<Enum extends string>(validValues: Enum[]) {
...
+ * const [color, setColor] = useQueryState(
+ *   'color',
+ *    parseAsStringConst(colors) // pass a readonly list of allowed values
+ *      .withDefault("red")
+ * )
+ * ```
+ *
+ * @param validValues The values you want to accept
+ */
+export function parseAsStringConst<Const extends string>(
```
franky47, line 251: **suggestion**: The TypeScript name for those is [literal](https://www.typescriptlang.org/docs/handbook/literal-types.html?azure-portal=true), so the parser could be named `parseAsStringLiteral`.

I thought of suggesting dropping the `String` in the name to allow using other literal types like numbers:

```ts
const diceParser = parseAsLiteral([1, 2, 3, 4, 5, 6] as const)
```

But this would require a different type of parsing to convert the query string value into the right type.

I wonder if a generic approach could be done to literal-ify the existing parsers (`parseAsString`, `parseAsInteger`, `parseAsFloat`, `parseAsBoolean` though this last one makes little sense I think).

packages/nuqs/src/parsers.ts
```diff
@@ -191,7 +191,7 @@ export const parseAsIsoDateTime = createParser({
 
 /**
  * String-based enums provide better type-safety for known sets of values.
- * You will need to pass the stringEnum function a list of your enum values
+ * You will need to pass the parseAsStringEnum function a list of your enum values
```
franky47, line 194: **praise** Thanks for catching that one and updating the docs!

franky47 (changes requested):
**praise**: That's a good idea, thanks!

I would only suggest using a name that maps directly with TypeScript: literals (see comment).

packages/nuqs/src/parsers.ts
timheerwagen: Thanks for the tip, unfortunately I'm not very familiar with the Typescript names. :)

I have renamed the function and it now accepts string or number literals.

packages/nuqs/src/parsers.ts
```diff
@@ -241,25 +241,25 @@ export function parseAsStringEnum<Enum extends string>(validValues: Enum[]) {
...
-export function parseAsStringConst<Const extends string>(
-  validValues: readonly Const[]
+export function parseAsLiteral<Literal extends string | number>(
+  validValues: readonly Literal[]
 ) {
   return createParser({
     parse: (query: string) => {
-      const asConst = query as unknown as Const
+      const asConst = query as unknown as Literal
       if (validValues.includes(asConst)) {
```
franky47, line 257: This is what I meant by having to extend the existing numeric parsers: using number literals here won't work as the array [1, 2, 3] never contains the value "2", it has to be parsed first.

We could hack our way out by detecting if the first element of the validValues is a number, then pass it through `parseFloat`, which handles all cases for numeric literals.

franky47 (changes requested):

packages/nuqs/src/parsers.ts
timheerwagen: Maybe it would be enough to provide 2 functions, that looks too much like a hack to me? `parseAsStringLiteral` and `parseAsNumberLiteral`

franky47: Sounds like a good plan indeed.

timheerwagen: Please take a look at this.

By the way, I can't think of any other uses for literals in query than String and Number. So a generic attempt would not really achieve much. You can always create your own parser for special use-cases.

README.md
```diff
@@ -130,12 +131,21 @@ const [direction, setDirection] = useQueryState(
...
+    .withDefault('red')
+)
+
+// Literals (number-based only)
+const diceSides = [1, 2, 3, 4, 5, 6] as const
+
+const [side, setSide] = useQueryState(
+  'side',
+  parseAsNumberLiteral(diceSides) // pass a readonly list of allowed values
     .withDefault('red')
```
franky47, line 149: **suggestion**: Update the doc to include a valid default value, like 4: https://xkcd.com/221/

packages/nuqs/src/parsers.ts
franky47: The only other allowed literal type is boolean, and we already have a parser for that one, so this is perfect, thanks!

franky47 (changes requested):
Last little change on the docs and it looks great.

README.md
timheerwagen: Oops, I must have overlooked that.

franky47 (approved):

PR #493, opened 2024-02-10
feat: Add "clear on default" option

### LINKED ISSUE #320: When updating value, remove from URL if new value == defaultValue
could be done [here](https://github.com/47ng/next-usequerystate/blob/f17d9a24227c2fc36bcd1aa2c3214187a7393432/src/useQueryState.ts#L253)

### LINKED ISSUE #488: Setting default value again should remove the query param from the url
## Context

1.16.1

Are you using:

- ❌ The app router
- ✅ The pages router
- ❌ The `basePath` option in your Next.js config
- ❌ The experimental `windowHistorySupport` flag in your Next.js config

## Description

Setting the default value again should remove the query param the same way as by setting it to null.

## Reproduction

1. Create a useQueryState with a default value
2. Update the state with a different value than the default value
3. Set the default value again
4. Query param is still in the url

 -->

franky47:
Adds an option to remove the key-value pair when setting a state to the default value. Default behaviour remains to keep it explicitly set in the URL, for backwards-compatibility when defaults change.

Closes #320.
Closes #488.
Closes #490.

### DIFF
--- a/packages/docs/content/docs/basic-usage.mdx
+++ b/packages/docs/content/docs/basic-usage.mdx
@@ -98,7 +98,7 @@ const clearCount = () => setCount(null) // Remove query from the URL
 
 <Callout title="Note">
   The default value is internal to React, it will **not** be written to the
-  URL.
+  URL _unless you set it explicitly_.
 </Callout>
 
 <Callout title="Tip">
--- a/packages/docs/content/docs/options.mdx
+++ b/packages/docs/content/docs/options.mdx
@@ -36,7 +36,7 @@ Call-level options will override hook level options.
 By default, state updates are done by replacing the current history entry with
 the updated query when state changes.
 
-You can see this as a sort of `git squash`, where all state-changing
+You can see this as a sort of `git squash{:shell}`, where all state-changing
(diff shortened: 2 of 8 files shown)


zbeyens (approved):
Looks good!

franky47 (commented):
Other than the naming of the option itself (which is hard to convey while keeping it short), one nitpick I may have with this implementation is that it's possible to specify `clearOnDefault` while not having set a default value, for which it will not do anything.

```ts
useQueryState('foo', parseAsInt.withOptions({
  clearOnDefault: true
})
```

But then again, the intent is clear, and the alternative would require a large refactor of the type system. This will probably happen in v2.

rwieruch:
Good addition to the package! FYI [use-query-params](https://www.npmjs.com/package/use-query-params#queryparamprovider) which I always used for SPAs and Next Page Router has this option as well. There it's called ** removeDefaultsFromUrl**.

franky47:
@rwieruch good catch, thanks! I'll update the docs to make the connection (since it's been released into stable, it's a bit late to change the name).

PR #504, opened 2024-02-22
fix: Equality check for non-primitives in clearOnDefault

franky47:
As reported by @Kavan72 in https://github.com/47ng/nuqs/discussions/462#discussioncomment-8554053

Testing only for referential equality is not enough for types like arrays and objects.

This introduces a new property of the parsers to specify an equality function to use when comparing against the default value.

- [x] Add jsdoc to the Parser type to reflect when to specify the `eq` property.

### DIFF
--- a/packages/e2e/cypress/e2e/clearOnDefault.cy.js
+++ b/packages/e2e/cypress/e2e/clearOnDefault.cy.js
@@ -1,7 +1,9 @@
 /// <reference types="cypress" />
 
 it('Clears the URL when setting the default value when `clearOnDefault` is used', () => {
-  cy.visit('/app/clearOnDefault?a=a&b=b')
+  cy.visit(
+    '/app/clearOnDefault?a=a&b=b&array=1,2,3&json-ref={"egg":"spam"}&json-new={"egg":"spam"}'
+  )
   cy.contains('#hydration-marker', 'hydrated').should('be.hidden')
   cy.get('button').click()
   cy.location('search').should('eq', '?a=')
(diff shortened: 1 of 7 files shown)


Kavan72:
any update here ?

Kavan72:
hey @franky47 is there any update on this ?

franky47:
@Kavan72 sorry I'm on holiday with sick wife and kids, kind of hard to free up time to work on this at the moment, hopefully I can merge this next week.

Kavan72:
@franky47 Great, take your time and have good weekends 😊

Kavan72:
@franky47 I think one edge case is still missing here. Suppose my URL looks like this:

`/example/?status=QUALIFIED,CONFIRMED,PENDING` 

and I'm changing this status with multiple checkboxes. Users can select any checkbox without any specific order. For example, if a user selects 'CONFIRMED' first, then 'PENDING', the URL would look like this:

`/example/?status=CONFIRMED,PENDING,QUALIFIED` 

In this case, the state remains the same but clearOnDefault will not be triggered due to checking the data with index using the `eq` method:

`a.every((value, index) => itemEq(value, b[index]!))`: 

here 

```
a: ['QUALIFIED', 'CONFIRMED', 'PENDING']
b: ['CONFIRMED', 'PENDING', 'QUALIFIED']
```
and this condition is not met because of `b[index]!`. Therefore, the condition is not the same.

franky47:
For this, I'd say you could write your own version of `parseAsArrayOf` and specify your own logic in the `eq` property.

In some cases, the order matters, in some other it doesn't.

PR #573, opened 2024-06-11
test: Add a case for (ab)using the cache for page/layout params

franky47:
Related to #572.


franky47:
Closing as this was mostly a thought experiment (and using CI to validate whether it could work). Definitely a hack, I would not recommend using it in production as it goes around co-location principles.

Issue #566, opened 2024-05-24
React Compiler / React 19


tordans:
I was experimenting with getting our app https://github.com/FixMyBerlin/atlas-app/pull/111 build with the preview of the React Compiler.

When I run `npm run build` I get a few errors that mention _nuqs_:

```
> atlas-app@0.2.0 build
> blitz build

Loaded env from /Users/fmc/Development/atlas-app/.env.production.local
Loaded env from /Users/fmc/Development/atlas-app/.env
✔ Next.js was successfully patched with a React Suspense fix
✔ Routes manifest was successfully generated
  ▲ Next.js 14.3.0-canary.80
  - Environments: .env.production.local, .env
  - Experiments (use with caution):
    · typedRoutes
    · reactCompiler
    · instrumentationHook

   Creating an optimized production build ...
Failed to compile.

./node_modules/nuqs/dist/index.js
Attempted import error: 'useRouter' is not exported from 'next/navigation.js' (imported as 'useRouter').

Import trace for requested module:
./node_modules/nuqs/dist/index.js
./src/app/regionen/[regionSlug]/_hooks/useQueryState/useMapParam.ts
./src/app/regionen/[regionSlug]/_hooks/useQueryState/useCategoriesConfig/migrations/0001_from_middleware.ts
./src/app/regionen/[regionSlug]/_hooks/useQueryState/useCategoriesConfig/migrations/index.ts
./src/app/regionen/[regionSlug]/_hooks/useQueryState/useCategoriesConfig/migrateUrl.ts

./node_modules/nuqs/dist/index.js
Attempted import error: 'useSearchParams' is not exported from 'next/navigation.js' (imported as 'useSearchParams').

Import trace for requested module:
./node_modules/nuqs/dist/index.js
./src/app/regionen/[regionSlug]/_hooks/useQueryState/useMapParam.ts
./src/app/regionen/[regionSlug]/_hooks/useQueryState/useCategoriesConfig/migrations/0001_from_middleware.ts
./src/app/regionen/[regionSlug]/_hooks/useQueryState/useCategoriesConfig/migrations/index.ts
./src/app/regionen/[regionSlug]/_hooks/useQueryState/useCategoriesConfig/migrateUrl.ts

./node_modules/nuqs/dist/index.js
Attempted import error: 'useRouter' is not exported from 'next/navigation.js' (imported as 'useRouter').

Import trace for requested module:
./node_modules/nuqs/dist/index.js
[...]

PR #581, opened 2024-06-27
fix: next/router causes issue with react compiler

tylerlaws0n:
I came across this issue as well when trying out the react compiler: https://github.com/47ng/nuqs/issues/566

What I tried to debug and fix:

1. Setup eslint in `packages/nuqs` and check for violations of the [rules of react](https://react.dev/reference/rules)
    I tried this because in theory, the react compiler should only result in a break if these rules are being violated, from what I understand of it so far. This however, did not seem to call out any issues that would result in the build errors. I'll open a second PR to show the addition of the lint setup, but the two changesets can be considered separately.

2. When running `next dev` locally I noticed the following errors upon request of a page:
image
This led me to find the import to `next/router`. This was just an import for the type of the router, so it seems very safe to change over to `next/navigation` for support with the app router. FWIW I only tested the changes on a project with the app router, but I can't think of a way this would break anything unless the call to `method.call` broke somehow.

Happy to make changes/adjustments as necessary with this, but hopefully this just fixes the issue :)

### DIFF
--- a/packages/nuqs/src/update-queue.ts
+++ b/packages/nuqs/src/update-queue.ts
@@ -1,4 +1,3 @@
-import type { NextRouter } from 'next/router'
 import { debug } from './debug'
 import type { Options, Router } from './defs'
 import { error } from './errors'
@@ -123,7 +122,7 @@ declare global {
   interface Window {
     next?: {
       version: string
-      router?: NextRouter & {
+      router?: Router & {
         state: {
           asPath: string
         }
@@ -170,7 +169,7 @@ function flushUpdateQueue(router: Router): [URLSearchParams, null | unknown] {
       debug('[nuqs queue (pages)] Updating url: %s', url)
       const method =
         options.history === 'push' ? nextRouter.push : nextRouter.replace
-      method.call(nextRouter, url, url, {
+      method.call(nextRouter, url, {
         scroll: options.scroll,
         shallow: options.shallow
       })


franky47:
The issue in #566 has to do with the import of `next/navigation.js`, where the compiler (or something else in the module resolution) doesn't like the explicit file extension. That fix had been put in place in https://github.com/47ng/nuqs/discussions/352#discussioncomment-6997334, to fix some incompatibility with tsconfig settings.

To be fair, a fix for the React compiler would belong better in the v2 branch, which will target Next.js 15 and React 19 exclusively.

Coming back to your changes, the import from `next/router` was correct: it's a type-only import, that augments the internal `next` global property where we can access the router outside of a hook context (in the URL update queue). This isn't something the React compiler will be able to memoize, but it doesn't violate the rules of hooks either, as we're using the 

Now I agree that the type can be misleading: in pages under the `pages` directory, the `window.next.router` will be an instance of NextRouter (from `next/router`, the traditional pages router). But in pages under the `app` directory, `window.next.router` will be an instance of the app router.

So there isn't a correct way to type this, but since Next.js does a full page reload when crossing those boundaries (if both routers are used in a single app), we're guaranteed to only have the "correct" router when accessing to update the URL. 

This hack is needed only in the pages router, for backwards compatibility with older versions of Next.js, as the router exposed from `next/navigation`, when imported in the pages router, is a stub file with some features missing (support for basePath, dynamic routes / asPath etc).

tylerlaws0n:
@franky47 understood! Thanks for the quick and thorough response here. 

I missed the v2 when looking through, so thanks for mentioning that. Looks like that will resolve everything that I've run up against so far after testing that locally. Thanks for the work on that!

PR #582, opened 2024-06-27
chore: Setup eslint for packages/nuqs

tylerlaws0n:
Sharing the lint setup I created when debugging #566 in case linting is desired going forward.

This is mostly the default setup from `npm init @eslint/config@latest` with eslint 8, react, and ts. I added eslint-plugin-react-hooks & eslint-plugin-react-compiler. 

I disabled the following rules because they throw errors in the current codebase that I didn't find particularly useful here:

```js
rules: {
  '@typescript-eslint/no-explicit-any': 'off',
  '@typescript-eslint/ban-ts-comment': 'off',
  'no-useless-escape': 'off'
}
```

I also disabled eslint entirely in the `*.test-d.ts` files because those threw many errors with the rules of hooks and if the current usage of those files works fine for testing, I don't see any reason to lint there (at least for the rules of hooks).

Lastly, there are just a few warnings related to dependency arrays of the current hook usages:

image

These can be fixed very easily, but I will wait for feedback on how y'all would like to fix them (or not fix them) before making further changes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
   },
   "scripts": {
     "dev": "turbo run dev",
+    "lint": "turbo run lint",
     "build": "turbo run build",
     "test": "turbo run test",
     "prepare": "husky"
--- a/packages/nuqs/eslint.config.js
+++ b/packages/nuqs/eslint.config.js
@@ -1,0 +1,39 @@
+import globals from 'globals'
+import pluginJs from '@eslint/js'
+import tseslint from 'typescript-eslint'
+import pluginReactConfig from 'eslint-plugin-react/configs/recommended.js'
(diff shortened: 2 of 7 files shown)


tylerlaws0n:
@franky47 should I close this? I'm guessing since it's been left here, you're likely not interested which is np for me, I just like to close things to keep my PR tabs clean if they're not going anywhere.

franky47:
Yeah I think we can close this. My experience with eslint has always been negative, though I see the value in a linter for community projects.

My eyes are on Biome to replace Prettier & linger steps, but last time I tried it it wasn't quite ready (especially regarding Tailwind support in the docs app).

PR #588, opened 2024-07-25
fix: localStorage availability check update

iuriizaporozhets:
When in Safari set Block All cookies then local storage also will not work. But window.localStorage object itself will be available and when we will try set or get something from local storage we will get `SecurityError: The operation is insecure.` error and it is crashing the application for users with disabled cookies and local storage.

### DIFF
--- a/packages/nuqs/src/utils.ts
+++ b/packages/nuqs/src/utils.ts
@@ -38,3 +38,22 @@ export function getDefaultThrottle() {
     return 320
   }
 }
+
+/**
+ * Check if localStorage is available
+ */
+export function isLocalStorageAvailable(){
+  try {
+    const test = 'test';
+
+    window.localStorage.setItem(test, test);
+
+    const isValueAvailable = window.localStorage.getItem(test) === test;
+
+    window.localStorage.removeItem(test);
+
+    return isValueAvailable;
+  } catch (_) {
+    return false;
+  }
+};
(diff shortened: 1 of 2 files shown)


packages/nuqs/src/utils.ts
```diff
@@ -38,3 +38,22 @@ export function getDefaultThrottle() {
...
   }
 }
+
+/**
+ * Check if localStorage is available
+ */
+export function isLocalStorageAvailable(){
+  try {
+    const test = 'test';
+
```
franky47, line 48: **formatting**: could you remove the extra empty lines between the code please?

packages/nuqs/src/utils.ts
```diff
@@ -38,3 +38,22 @@ export function getDefaultThrottle() {
     return 320
   }
 }
+
+/**
+ * Check if localStorage is available
+ */
+export function isLocalStorageAvailable(){
+  try {
+    const test = 'test';
```
franky47, line 47: **suggestion**: make the key less prone to collisions, for example by marking it specifically for nuqs: `nuqs-localStorage-test`.

franky47 (changes requested):
Thanks for this!

I see an issue with the current way it's implemented: if a client has a localStorage value under the `test` key, it would get cleared by what should otherwise be an idempotent function.

One alternative would be to read the value before attempting a write, or use a more specifically-named key, which won't cause collisions.

packages/nuqs/src/utils.ts
iuriizaporozhets: @franky47 updated to `nuqs-localStorage-test` as suggested, thanks

franky47 (approved):
LGTM, thanks!

PR #611, opened 2024-08-14
docs: renamed hello state to name in hero section demo

johurul-haque:
The demo in the hero section contains the state for handling the input field. The state variables for that are `hello` and `setHello`; for storing the name of the user it would be better to rename them to `name` and `setName`.

### DIFF
--- a/packages/docs/src/app/(pages)/_landing/demo.client.tsx
+++ b/packages/docs/src/app/(pages)/_landing/demo.client.tsx
@@ -3,7 +3,7 @@
 import { parseAsInteger, useQueryState } from 'nuqs'
 
 export function Demo() {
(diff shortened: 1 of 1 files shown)


franky47:
Thanks for the suggestion!

This is a good occasion to highlight something important with URL state: migrations like this can (and will) break existing URLs.

I don't mind merging the change in variable name, but the query key should remain the same to ensure existing URLs are still valid. But doing so, the variable name would be decoupled from its key.

So what's the best way to handle such migrations? One way is to rewrite the query string with [middleware](https://nextjs.org/docs/app/building-your-application/routing/middleware). It may break locality of behaviour, but it also centralises the migration code.

franky47:
Closing as not planned: this would need a migration step, which we (currently) don't have an API for.

See also: #665.

PR #638, opened 2024-09-16
feat: Sort query string keys

hugotiger:
To avoid SEO issues from duplicate content this PR adds sorting of the query parameters based on their keys. That way the URL should be more predictable and canonicals won't be needed as often. 

I went with using the native `sort` method on the `URLSearchParams` class to align it with existing standards. 

As long as no user has depended on the order of the query string this PR shouldn't be breaking.


franky47:
Thanks! Could you rebase it on top of the `v2/release` branch, and change the PR target branch to that please?

franky47:
Also, as an FYI, this change may be unexpected for most people: https://x.com/fortysevenfx/status/1835384141808021801

I'll play with this PR once it's ready, and we'll see if it makes sense to merge it as-is or to provide some other kind of API to sort the keys in a one-shot way, or to provide this behaviour as an option.

hugotiger:
> Also, as an FYI, this change may be unexpected for most people: https://x.com/fortysevenfx/status/1835384141808021801
> 
> I'll play with this PR once it's ready, and we'll see if it makes sense to merge it as-is or to provide some other kind of API to sort the keys in a one-shot way, or to provide this behaviour as an option.

I've 

> Also, as an FYI, this change may be unexpected for most people: https://x.com/fortysevenfx/status/1835384141808021801
> 
> I'll play with this PR once it's ready, and we'll see if it makes sense to merge it as-is or to provide some other kind of API to sort the keys in a one-shot way, or to provide this behaviour as an option.

A rough first version is ready now. It's one of my first time contributing to an OSS-project so sorry if I'm missing something. Let me know and I'll fix it 😄 

I understand this change might be considered somewhat controversial, especially regarding backward compatibility. Like you're saying, it might be good to have an option to enable/disable it. 

However, I believe the default behavior should remain sorted. While it may not align with everyone’s expectations, it provides significant SEO benefits, which would likely be valuable to a majority of users. Similar to how this library has some handy defaults (replace, reset on null etc.), I believe sorting should be the default as it’s what would be the most practical.

packages/nuqs/src/url-encoding.ts
```diff
@@ -2,8 +2,13 @@ export function renderQueryString(search: URLSearchParams) {
...
     return ''
   }
+
+  const clonedSearch = new URLSearchParams(search)
+  clonedSearch.sort()
+  const entries = clonedSearch.entries()
+
   const query: string[] = []
-  for (const [key, value] of search.entries()) {
+  for (const [key, value] of entries) {
```
tylerlaws0n, line 11: `URLSearchParams` is iterable on its own which is neat

Screenshot 2024-09-23 at 7 03 53 PM

A SO answer from a self-described v8 dev on the cost of `Object.entries()` https://stackoverflow.com/a/64911507

I am assuming URLSearchParams proxies to Object.entries or is implemented the same way. This feels a bit pedantic, but if encoding happens frequently and with a large amount of parameters, I could imagine this mattering.

tylerlaws0n: I tried a benchmark and my theory of this helping was dead wrong lol sorry 

benchmark: https://jsben.ch/qndxn

still is neat to know imo, but please disregard other than the syntax difference

franky47: I suppose it's implemented the same way internally.

The main advantage of iterating directly on the URLSearchParams is skipping an extra `entries` variable, so saving memory and a few bytes of bundle size.

franky47 (changes requested):
Thinking about this, we should probably mutate the URLSearchParams passed in argument rather than copying, as it will be returned by the Promise of useQueryState(s) state updater functions, and it should reflect the order.

To keep the encoding function pure, the sort call should then happen after the search params have been populated, but before encoding.

This could also allow placing this behaviour behind an option. A good place for it would be the `flushUpdateQueue` function in `update-queue.ts`.

Issue #641, opened 2024-09-18
`clearOnDefault` doesn't work with `createSerializer`


hugotiger:
## Context

What's your version of `nuqs`?

```
"nuqs": "1.19.1"
```

Next.js information (obtained by running `next info`):

```
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 23.3.0: Wed Dec 20 21:30:59 PST 2023; root:xnu-10002.81.5~7/RELEASE_ARM64_T6030
  Available memory (MB): 36864
  Available CPU cores: 11
Binaries:
  Node: 20.11.1
  npm: 10.2.4
  Yarn: 1.18.0
  pnpm: N/A
Relevant Packages:
  next: 14.2.11 // There is a newer version (14.2.12) available, upgrade recommended! 
  eslint-config-next: N/A
  react: 18.2.0
  react-dom: 18.2.0
  typescript: 5.5.4
Next.js Config:
  output: N/A
 ⚠ There is a newer version (14.2.12) available, upgrade recommended! 
   Please try the latest canary version (`npm install next@canary`) to confirm the issue still exists before creating a new issue.
   Read more - https://nextjs.org/docs/messages/opening-an-issue
```

Are you using:

- ✅ The app router
- ❌ The pages router
- ❌ The `basePath` option in your Next.js config
- ❌ The experimental `windowHistorySupport` flag in your Next.js config

## Description
The `clearOnDefault` option does not seem to work when using `createSerializer`. Seeing as TS was happy I expected it to work. I love this feature from the hooks so would be amazing getting it to work!

Let me know if I should attempt adding/fixing this and I'll take a stab at it :) 

## Reproduction

 I skipped creating a CodeSandbox playground since it's very easy to explain. Let me know if I still should create one 😄 

```ts
const serialize = createSerializer({
    int: parseAsInteger.withOptions({ clearOnDefault: true }).withDefault(0)
})
const result = serialize({ int: 0 })
//    ^ Result is `?int=0`
```

franky47:
Ah yes, that would make sense for the serializer to follow this setting. If you want to open a PR, that'd be great, thanks!

hugotiger:
> Ah yes, that would make sense for the serializer to follow this setting. If you want to open a PR, that'd be great, thanks!

[...]

PR #642, opened 2024-09-19
fix: Serialization of default values in `createSerializer`

hugotiger:
When using parsers in `createSerializer` it would not respect the `clearOnDefault` option properly when serializing, thus adding key-value pairs to the string that shouldn't be there.

An example:
```ts
const serialize = createSerializer({
    int: parseAsInteger.withOptions({ clearOnDefault: true }).withDefault(0)
})
const result = serialize({ int: 0 })
//    ^ Result is `?int=0` instead of empty string
```

Closes https://github.com/47ng/nuqs/issues/641.

### DIFF
--- a/packages/nuqs/src/serializer.test.ts
+++ b/packages/nuqs/src/serializer.test.ts
@@ -1,5 +1,11 @@
 import { describe, expect, test } from 'vitest'
-import { parseAsBoolean, parseAsInteger, parseAsString } from './parsers'
+import {
+  parseAsBoolean,
+  parseAsInteger,
+  parseAsString,
+  parseAsJson,
+  parseAsArrayOf
+} from './parsers'
 import { createSerializer } from './serializer'
 
 const parsers = {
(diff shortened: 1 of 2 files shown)


franky47:
Thanks! Could you change back the base branch to next for this one? It's not a breaking change, so it can be published on the v1 release line.

I pushed a fix for the types to let parsers know they can have a default value.

franky47 (changes requested):
One test is failing, see https://github.com/47ng/nuqs/actions/runs/10936702232/job/30378534527#step:7:90

hugotiger:
> One test is failing, see https://github.com/47ng/nuqs/actions/runs/10936702232/job/30378534527#step:7:90

I'm not able to reproduce it locally 🤔 Could it be some caching issue in the action? For context when I run `pnpm run test` it's all green.

hugotiger:
> > One test is failing, see https://github.com/47ng/nuqs/actions/runs/10936702232/job/30378534527#step:7:90
> 
> I'm not able to reproduce it locally 🤔 Could it be some caching issue in the action? For context when I run `pnpm run test` it's all green.

I checked again now and it seems as if the CI was running on some previous commit. Could you do a rerun?

franky47 (approved):

franky47:
Thanks!

PR #647, opened 2024-09-25
fix(serializer): Allow serializers to operate on derived `UrlSearchParams` classes

neefrehman:
Follow-up to https://github.com/47ng/nuqs/discussions/646, that allows us to use the response from Next's `useSearchParams`, for example, as the `base` argument to a `serializer`. To do this we create a copy of `UrlSearchParams` in this code branch.

### DIFF
--- a/packages/nuqs/src/serializer.ts
+++ b/packages/nuqs/src/serializer.ts
@@ -72,6 +72,9 @@ function splitBase(base: Base) {
   } else if (base instanceof URLSearchParams) {
     return ['', new URLSearchParams(base)] as const // Operate on a copy of URLSearchParams, as derived classes may restrict its allowed methods
   } else {
-    return [base.origin + base.pathname, new URLSearchParams(base.searchParams)] as const
+    return [
+      base.origin + base.pathname,
+      new URLSearchParams(base.searchParams)
+    ] as const
   }
 }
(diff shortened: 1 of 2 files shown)


packages/nuqs/src/serializer.ts
```diff
@@ -70,7 +70,7 @@ function splitBase(base: Base) {
     const [path = '', search] = base.split('?')
     return [path, new URLSearchParams(search)] as const
   } else if (base instanceof URLSearchParams) {
-    return ['', base] as const
+    return ['', new URLSearchParams(base)] as const // Operate on a copy of URLSearchParams, as derived classes may restrict its allowed methods
   } else {
     return [base.origin + base.pathname, base.searchParams] as const
```
franky47, line 75: **suggestion**: Could we create a new URLSearchParams object on `base.searchParams`, to avoid mutating the URL object passed as `base`?

franky47 (changes requested):
Thanks!

Could you please add a test to verify that the base parameter is no longer mutated?

neefrehman:
@franky47 thanks for the review! I've just implemented the change and test

franky47 (approved):
**praise**: LGTM, thanks!

franky47:
Sorry for the close/reopen, my CI workflow wasn't picking subsequent commits from external contributors (should be fixed in #649).

PR #656, opened 2024-09-27
Fix: Run pkg.pr.new only if there are changes

### LINKED ISSUE #655: Don't run `pkg.pr.new` if there aren't changes to the `nuqs` package
[pkg.pr.new](https://pkg.pr.new) is a service by StackBlitz to deploy previews of NPM packages without polluting the NPM registry (it's pretty cool).

At the moment the PR preview workflow ([`pkg.pr.new.yml`](https://github.com/47ng/nuqs/tree/next/.github/workflows/pkg.pr.new.yml)) runs on any PR.

This is kind of wasteful on pkg.pr.new's servers and resources: we should abort if the PR introduces no changes to what's actually being deployed.

PRs welcome!

chbg:
Added a conditional check in pkg.pr.new.yml to prevent unnecessary runs when no changes are detected in the nuqs package.

Closes #655.

### DIFF
--- a/.github/workflows/pkg.pr.new.yml
+++ b/.github/workflows/pkg.pr.new.yml
@@ -13,10 +13,11 @@ jobs:
     steps:
       - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332
       - name: Check for changes to nuqs package
+        id: check-for-changes
         run: |
           if git diff --quiet HEAD^ HEAD ./packages/nuqs; then
             echo "No changes to nuqs package, skipping preview deployment."
-            exit 0
+            echo "skip=true" >> $GITHUB_OUTPUT
           fi
       - uses: pnpm/action-setup@fe02b34f77f8bc703788d5817da081398fad5dd2
       - uses: actions/setup-node@1e60f620b9541d16bece96c5465dc8ee9832be0b
--- a/.github/workflows/pkg.pr.new.yml
+++ b/.github/workflows/pkg.pr.new.yml
@@ -17,7 +17,7 @@ jobs:
       - name: Check for changes to nuqs package
         id: check-for-changes
         run: |
-          if git diff --quiet ${{ github.event.pull_request.base.sha }} ${{ github.event.pull_request.head.sha }} ./packages/nuqs; then
+          if git diff --quiet HEAD^ HEAD ./packages/nuqs; then
             echo "No changes to nuqs package, skipping preview deployment."
             echo "skip=true" >> $GITHUB_OUTPUT
           fi
--- a/.github/workflows/pkg.pr.new.yml
+++ b/.github/workflows/pkg.pr.new.yml
@@ -12,6 +12,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332
+      - name: Check for changes to nuqs package
+        run: |
+          if git diff --quiet HEAD^ HEAD ./packages/nuqs; then
+            echo "No changes to nuqs package, skipping preview deployment."
+            exit 0
+          fi
       - uses: pnpm/action-setup@fe02b34f77f8bc703788d5817da081398fad5dd2
       - uses: actions/setup-node@1e60f620b9541d16bece96c5465dc8ee9832be0b
         with:
--- a/.github/workflows/pkg.pr.new.yml
+++ b/.github/workflows/pkg.pr.new.yml
@@ -15,7 +15,7 @@ jobs:
       - name: Check for changes to nuqs package
         id: check-for-changes
         run: |
-          if git diff --quiet HEAD^ HEAD ./packages/nuqs; then
+          if git diff --quiet ${{ github.event.pull_request.base.sha }} ${{ github.event.pull_request.head.sha }} ./packages/nuqs; then
             echo "No changes to nuqs package, skipping preview deployment."
             echo "skip=true" >> $GITHUB_OUTPUT
           fi
--- a/.github/workflows/pkg.pr.new.yml
+++ b/.github/workflows/pkg.pr.new.yml
@@ -12,6 +12,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332
+        with:
+          fetch-depth: 2
       - name: Check for changes to nuqs package
         id: check-for-changes
         run: |
--- a/.github/workflows/pkg.pr.new.yml
+++ b/.github/workflows/pkg.pr.new.yml
@@ -32,5 +32,5 @@ jobs:
         run: pnpm pkg set version=0.0.0-preview.${{ github.event.pull_request.head.sha }}
         working-directory: packages/nuqs
       - name: Publish to pkg.pr.new
-      - if: steps.check-for-changes.outputs.skip != 'true'
+        if: steps.check-for-changes.outputs.skip != 'true'
         run: pnpx pkg-pr-new publish --compact './packages/nuqs'
--- a/.github/workflows/pkg.pr.new.yml
+++ b/.github/workflows/pkg.pr.new.yml
@@ -33,5 +33,4 @@ jobs:
         working-directory: packages/nuqs
       - name: Publish to pkg.pr.new
       - if: steps.check-for-changes.outputs.skip != 'true'
-        
         run: pnpx pkg-pr-new publish --compact './packages/nuqs'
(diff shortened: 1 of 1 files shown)


.github/workflows/pkg.pr.new.yml
```diff
@@ -12,6 +12,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332
+      - name: Check for changes to nuqs package
+        run: |
+          if git diff --quiet HEAD^ HEAD ./packages/nuqs; then
+            echo "No changes to nuqs package, skipping preview deployment."
+            exit 0
+          fi
```
franky47, line 20: **suggestion**: An exit 0 code will only exit this particular step, and the rest of the workflow will keep going.

Rather than failing here (we still want the workflow to succeed, but skip build & deployment steps), we can set an output for this step and skip subsequent steps based on that output.

Also, the check on Git only accounts for the latest commit in the PR. It would be preferable to check the whole commit range from base to HEAD.

```suggestion
        id: check-for-changes
        run: |
          if git diff --quiet HEAD^ HEAD ./packages/nuqs; then
            echo "No changes to nuqs package, skipping preview deployment."
            echo "skip=true" >> $GITHUB_OUTPUT
          fi
```

.github/workflows/pkg.pr.new.yml
```diff
@@ -25,4 +31,4 @@ jobs:
         run: pnpm pkg set version=0.0.0-preview.${{ github.event.pull_request.head.sha }}
         working-directory: packages/nuqs
       - name: Publish to pkg.pr.new
-        run: pnpx pkg-pr-new publish --compact './packages/nuqs'
+        run: pnpx pkg-pr-new publish --compact './packages/nuqs'
```
franky47, line 34: **suggestion**: Can we keep the final newline here?

.github/workflows/pkg.pr.new.yml
```diff
@@ -25,4 +31,4 @@ jobs:
         run: pnpm pkg set version=0.0.0-preview.${{ github.event.pull_request.head.sha }}
         working-directory: packages/nuqs
       - name: Publish to pkg.pr.new
```
franky47, line 33: **suggestion**: Use the `skip` output from the check step:

```suggestion
      - name: Publish to pkg.pr.new
      - if: steps.check-for-changes.outputs.skip != 'true'
```

franky47 (changes requested):
Thanks! That issue was initially intended for Hacktoberfest, but there's no harm in getting it done early.

A couple of changes are needed:
- `exit 0` doesn't actually skip rest of the steps in the workflow
- Only the last commit in the PR is checked for changes

.github/workflows/pkg.pr.new.yml
```diff
@@ -25,4 +32,6 @@ jobs:
...
-        run: pnpx pkg-pr-new publish --compact './packages/nuqs'
+      - if: steps.check-for-changes.outputs.skip != 'true'
+        
+        run: pnpx pkg-pr-new publish --compact './packages/nuqs'
```
franky47, line 37: The missing newline I mentioned was the one at the end of the file, not between if and run.

```suggestion
        run: pnpx pkg-pr-new publish --compact './packages/nuqs'

```

.github/workflows/pkg.pr.new.yml
```diff
@@ -25,4 +32,6 @@ jobs:
         run: pnpm pkg set version=0.0.0-preview.${{ github.event.pull_request.head.sha }}
         working-directory: packages/nuqs
       - name: Publish to pkg.pr.new
-        run: pnpx pkg-pr-new publish --compact './packages/nuqs'
+      - if: steps.check-for-changes.outputs.skip != 'true'
```
franky47, line 35: Adding a `-` on the if causes it to become a new step, making the `- name` on the previous line a step without any action.

```suggestion
        if: steps.check-for-changes.outputs.skip != 'true'
```

franky47 (changes requested):

.github/workflows/pkg.pr.new.yml
franky47: Sorry I realized my suggestion was the one that included the `-`, that'll teach me to copy-paste without reviewing my own code!

.github/workflows/pkg.pr.new.yml
```diff
@@ -12,6 +12,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332
+      - name: Check for changes to nuqs package
+        id: check-for-changes
+        run: |
+          if git diff --quiet HEAD^ HEAD ./packages/nuqs; then
```
franky47, line 18: Now for the Git diff part, we could try this (though `github.event.pull_request.base.sha` might not be right in case we rebase the PR, TBC).

```suggestion
          if git diff --quiet ${{ github.event.pull_request.base.sha }} ${{ github.event.pull_request.head.sha }} ./packages/nuqs; then
```

franky47 (changes requested):

franky47 (approved):
Thanks! Let's see if it works now.

chbg:
> Thanks! Let's see if it works now.

Thanks for your guidance and patience. I really appreciate it.

.github/workflows/pkg.pr.new.yml
```diff
@@ -12,6 +12,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332
```
franky47, line 14: Nope, the base branch didn't work, it did resolve to the right SHA-1, but the commit object is not in the tree, that means our checkout step isn't deep enough. Might need to tweak some options [here](https://github.com/actions/checkout).

franky47 (changes requested):

.github/workflows/pkg.pr.new.yml
franky47: Hold on, your initial suggestion of `HEAD HEAD^` might have been correct, due to the temporary merge commit GitHub creates to run CI workflows on:
https://www.kenmuse.com/blog/the-many-shas-of-a-github-pull-request/

chbg: If it doesn't work, and we want to explore the checkout step, we could try: 

```
- uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332
  with:
    fetch-depth: 0  # This fetches all history for all branches and tags
```

Or: 

```
- uses: actions/checkout@v2
  with:
    fetch-depth: 0  # Fetches all history for all branches
    fetch-tags: true  # Fetches all tags
```

franky47: A fetch of 0 might be a bit overkill, if we fetch with a depth of 2, we can get the current HEAD (aka the temporary merge commit) and its first parent (aka the commit on the base branch where the PR would be merged).

So here:
```suggestion
      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332
        with:
          fetch-depth: 2
```

And back to `HEAD HEAD^` in the git diff.

.github/workflows/pkg.pr.new.yml
```diff
@@ -12,6 +12,13 @@ jobs:
...
+      - name: Check for changes to nuqs package
+        id: check-for-changes
+        run: |
+          if git diff --quiet ${{ github.event.pull_request.base.sha }} ${{ github.event.pull_request.head.sha }} ./packages/nuqs; then
```
franky47, line 18:
```suggestion
          if git diff --quiet HEAD HEAD^ ./packages/nuqs; then
```

franky47 (changes requested):

franky47 (approved):
Let's give it another try.

franky47:
It worked! Thanks for your patience too, TIL a few more things about Git and GHA 😄

chbg:
> It worked! Thanks for your patience too, TIL a few more things about Git and GHA 😄 

We did it! Onto the next one... 🤝

franky47:
Well if you feel like it, we could optimise this even more by splitting the workflow into two jobs:

1. Checkout with depth 2, doing the check via git, and emitting a job-level output for the skip signal
2. The original build + deploy, where the whole job would get skipped if not needed.

That saves setting up dependencies and building the package if we're not going to do anything with it.

chbg:
> Well if you feel like it, we could optimise this even more by splitting the workflow into two jobs:
> 
> 
> 
> 1. Checkout with depth 2, doing the check via git, and emitting a job-level output for the skip signal
> 
> 2. The original build + deploy (with a default checkout now), where the whole job would get skipped if not needed.
> 
> 
> 
> That saves setting up dependencies and building the package if we're not going to do anything with it.

I feel like it! Want me to just submit another PR when I'm done or do you want to create an issue and assign it to me?

franky47:
Go for it, no need for an issue.

chbg:
> Go for it, no need for an issue (maybe refer to the original #655 one for reference).

The PR is ready for 'ya [here](https://github.com/47ng/nuqs/pull/659).

PR #657, opened 2024-09-27
feat(serializer): allow `null` to clear all param values from base

neefrehman:
Follow-up to https://github.com/47ng/nuqs/discussions/653, this PR allows callers to pass `null` to a serializer instance to clear any search params from `base`.

### DIFF
--- a/packages/nuqs/src/serializer.test.ts
+++ b/packages/nuqs/src/serializer.test.ts
@@ -70,10 +70,15 @@ describe('serializer', () => {
     const result = serialize(url, { str: 'foo' })
     expect(result).toBe('https://example.com/path?bar=egg&str=foo')
   })
-  test('null deletes from base', () => {
+  test('null value deletes from base', () => {
     const serialize = createSerializer(parsers)
     const result = serialize('?str=bar&int=-1', { str: 'foo', int: null })
     expect(result).toBe('?str=foo')
+  })
+  test('null deletes all from base', () => {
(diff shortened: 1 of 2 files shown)


packages/nuqs/src/serializer.test.ts
```diff
@@ -70,11 +70,16 @@ describe('serializer', () => {
...
+  test('null value deletes from base', () => {
     const serialize = createSerializer(parsers)
     const result = serialize('?str=bar&int=-1', { str: 'foo', int: null })
     expect(result).toBe('?str=foo')
   })
+  test('null deletes all from base', () => {
+    const serialize = createSerializer(parsers)
+    const result = serialize('?str=bar&int=-1', null)
+    expect(result).toBe('')
+  })
```
franky47, line 82: **suggestion**: Could we add a test that ensures that other search params (not managed by the serializer) are kept in the output?

```suggestion
  })
  test('null keeps search params not managed by the serializer', () => {
    const serialize = createSerializer(parsers)
    const result = serialize('?str=foo&external=kept', null)
    expect(result).toBe('?external=kept')
  })
```

franky47 (approved):
**praise**: LGTM, thanks!

packages/nuqs/src/serializer.test.ts
neefrehman: Just done that, thanks!

PR #665, opened 2024-10-03
fix: Update query state key from 'hello' to 'name' in BasicUsageDemo

kzhgit:
This PR fixes a small issue in the BasicUsageDemo where the query state key was incorrectly set to 'hello'. The correct key should be 'name', as it more accurately represents the state being queried.

By making this adjustment, the code becomes clearer and aligns better with the expected functionality. I have tested the changes locally to ensure the demo behaves as intended with the updated key.

### DIFF
--- a/packages/docs/content/docs/demos.tsx
+++ b/packages/docs/content/docs/demos.tsx
@@ -48,7 +48,7 @@ function DemoContainer({ children, className, ...props }: DemoContainerProps) {
 }
 
 export function BasicUsageDemo() {
-  const [value, setValue] = useQueryState('hello', { defaultValue: '' })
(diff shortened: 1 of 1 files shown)


franky47:
This is a duplicate of #611, and my [comment](https://github.com/47ng/nuqs/pull/611#issuecomment-2288192213) there still stands: this would break existing URLs, and [cool URLs don't break](https://www.w3.org/Provider/Style/URI).

That being said, I would accept a change of the key if there was a middleware doing the translation from the old key to the new one.

franky47:
Closing as not planned: this would need a migration step, which we (currently) don't have an API for.

PR #677, opened 2024-10-11
chore: Sync up Next.js versions with PNPM catalogs

franky47:
Since catalogs can't be easily updated, the actual fix is in the workflows, where we install the Next.js version to both nuqs and e2e.


franky47:
Replaced by #678

PR #707, opened 2024-10-25
feat: Add parseAsISODate

tylersayshi:
follow up to https://github.com/47ng/nuqs/discussions/705


packages/docs/content/docs/parsers/built-in.mdx
```diff
@@ -183,12 +184,22 @@ The query string value will be the **value** of the enum, not its name
...
+<Suspense>
+  <DatetimeISOParserDemo/>
+</Suspense>
+
+### ISO 8601 Date
+
+```ts
+import { parseAsIsoDate } from 'nuqs'
+```
+
```
franky47, line 202: **suggestion**: Should we add a disclaimer that the returned Date object is rooted in UTC ?

packages/docs/content/docs/parsers/demos.tsx
```diff
@@ -290,10 +291,14 @@ export function DateParserDemo({
   )
 }
 
-export function DateISOParserDemo() {
+export function DatetimeISOParserDemo() {
   return <DateParserDemo queryKey="iso" parser={parseAsIsoDateTime} />
 }
 
+export function DateISOParserDemo() {
+  return <DateParserDemo queryKey="iso" parser={parseAsIsoDate} />
```
franky47, line 299: **suggestion**: The input inside the `<DateParserDemo>` component is currently harcoded to `type="datetime-local"`. For this parser, we're not interested in the time, so a `type="date"` would make more sense. We could add a prop to `DateParserDemo` to do that.

Also, binding to the same queryKey will cause issues, see the caveats section of the docs:
https://nuqs.47ng.com/docs/troubleshooting#caveats

packages/nuqs/src/parsers.test.ts
```diff
@@ -48,6 +49,16 @@ describe('parsers', () => {
...
   })
+  test('parseAsIsoDate', () => {
+    expect(parseAsIsoDate.parse('')).toBeNull()
+    expect(parseAsIsoDate.parse('not-a-date')).toBeNull()
+    const moment = '2020-01-01'
+    const ref = new Date(moment)
+    expect(parseAsIsoDate.parse(moment)).toStrictEqual(ref)
+    expect(parseAsIsoDate.parse(moment.slice(0, 10))).toStrictEqual(ref)
+    expect(parseAsIsoDate.serialize(ref).length).toBe(10)
+    expect(parseAsIsoDate.serialize(ref).length).not.toContain('T')
```
franky47, line 60: suggestion: We could check for equality with the expected format here.

```suggestion
    expect(parseAsIsoDate.serialize(ref)).toEqual(moment)
```

packages/nuqs/src/parsers.ts
```diff
@@ -215,6 +215,22 @@ export const parseAsIsoDateTime = createParser({
...
 })
 
+/**
+ * Querystring encoded as an ISO-8601 string (UTC)
+ * without the time zone offset, and returned as
+ * a Date object.
+ */
+export const parseAsIsoDate = createParser({
+  parse: v => {
+    const date = new Date(v + 'T00:00:00.000Z')
```
franky47, line 225: **suggestion**: Clip the input to 10 chars to remove time information from full dates. Eg:

`?date=2020-01-01T12:34:56.789Z` -> `?date=2020-01-01T00:00:00.000Z` rather than an invalid date.

```suggestion
    const date = new Date(v.slice(0, 10) + 'T00:00:00.000Z')
```

Do we need to append the time information? `new Date('2020-01-01')` already returns the correct UTC value of midnight for that day (tested in Firefox and Node.js)

packages/docs/content/docs/parsers/built-in.mdx
tylersayshi: Good call! Disclaimer added to the jsdoc and the docs here 👍

packages/docs/content/docs/parsers/demos.tsx
tylersayshi: nice! fixed both. Thanks for calling out the queryKey, that was just an oversight while copy-pasteing the existing docs.

packages/nuqs/src/parsers.test.ts
tylersayshi: nice, yeah this cleans up and is more comprehensive a test. Thanks

packages/nuqs/src/parsers.ts
tylersayshi: I think I just misread a timestamp while building this initially. I just tested the same and got the same result. 

Fixed to remove the extra `'T00:00:00.000Z'`

packages/docs/content/docs/parsers/demos.tsx
```diff
@@ -292,15 +294,29 @@ export function DateParserDemo({
 }
 
 export function DatetimeISOParserDemo() {
-  return <DateParserDemo queryKey="iso" parser={parseAsIsoDateTime} />
```
franky47: **suggestion**: Can we keep the `iso` key here? In order not to break existing URLs (I doubt many people would have linked to the parsers demos, but it's always a good exercise).

We could have:
- `iso`: full ISO-8601 (date + time)
- `date`: just the first 10 characters (date string)
- `ts`: unix timestamp ms

packages/nuqs/src/parsers.ts
```diff
@@ -219,10 +219,13 @@ export const parseAsIsoDateTime = createParser({
  * Querystring encoded as an ISO-8601 string (UTC)
  * without the time zone offset, and returned as
  * a Date object.
+ *
+ * The Date is parsed without the time zone offset,
+ * making it at GMT 00:00:00 UTC.
```
franky47, line 224: **nitpick**: Can't be both GMT & UTC (even though they are equal in our case)

```suggestion
 * making it at 00:00:00 UTC.
```

franky47 (changes requested):
Thanks! I'd rather keep the `iso` query key for the datetime parser to avoid potential URL breakage.

I forgot I had Prettier disabled for MDX locally, I'll have to put that in the config. Not a big deal, we'll see if it creates formatting issues with the preview deployment.

tylersayshi:
> I forgot I had Prettier disabled for MDX locally, I'll have to put that in the config. Not a big deal, we'll see if it creates formatting issues with the preview deployment.

I had been formatting without save before then accidentally let it format. Would you like me to revert that?

franky47 (approved):
LGTM, thanks Tyler!

PR #716, opened 2024-10-28
doc: Update tips-tricks.mdx

gsaandy:


franky47:
Thanks for catching that!

PR #723, opened 2024-10-30
style: Use useSyncExternalStore in the React adapter

bartlangelaan:
This utilizes the `useSyncExternalStore` hook in the React adapter, which automatically takes care of rehydration errors.


franky47:
Thanks, wow this is so much cleaner! 

I had heard of useSyncExternalStore, but it didn't occur to me to use it here, it sounds like a perfect use-case. Great job!

bartlangelaan:
I don't have time to test it today, so if you can't wait and want to test it, feel free! :-)

franky47:
Regarding the emitter, it's use is to make sure the `searchParams` returned by the adapter hook is reactive: when the URL changes, no matter the reason (internal query state change, Link, back/fwd navigation), every hook needs to get the new value.

bartlangelaan:
I don't mind at all! I don't know when I will have time for this, so feel free!

franky47:
Closing this as we do use uSES for the optimistic state in core now, although we might migrate away from it to prefer useOptimistic when we start only supporting React ^19 (probably in nuqs@^3).

PR #724, opened 2024-10-30
feat: Async cache [experimental]

franky47:
To leverage Next.js 15's async searchParams prop and PPR, we can defer the parsing of search params to where they are actually used, rather than blocking rendering for top-level elements of the page.

Highly experimental, not sure how best to introduce it retro-compatibly with the existing cache (with sync/async overloads, but still top-level blocking PPR). So for now it's a fork.


franky47:
Closing as not planned, since there may be changes to the way the React `cache` function works that impacts the nuqs cache feature.

Source:
image

PR #742, opened 2024-11-05
Small typo in README

keul:


franky47:
This isn't a typo, it's a TypeScript non-null assertion operator that says "I know `getElementById` can return `null`, but we have the HTML right here and it's there, so mark it as non-null".

PR #745, opened 2024-11-06
Update basic-usage.mdx

destocot:
Wrapped c ?? 0 in paranthesis, doesn't work as expected otherwise


franky47 (approved):
Good catch, thanks!

PR #748, opened 2024-11-08
doc: tanstack-table refactor pagination example code, and add sorting, filtering, and playground

acharlop:
I'm working on this now for my own project so figured I'd help out and add documentation.

Please let me know if this is helpful and heading in the right direction and I'll continue to add the filtering and a playground.


franky47:
Thanks!

The idea behind community parsers is to highlight how to leverage custom parsers to represent data structures in the URL in a nice way. One thing you could try is to format the URL to only show the id and direction, like `?sort=name.asc` or `?sort=age:desc`.

The playground is an interactive way of configuring the parsers, variables here could include:
- The separator (`.`, `:`, `~` would be good choices, making sure to encode any instance of it in the id)
- The rendering of the direction: `asc | desc`, `a|d`, `^|v`

Does TanStack Table support sorting over multiple columns? Is this something we could support by combining it with `parseAsArrayOf`?

packages/docs/content/docs/parsers/community/tanstack-table.mdx
```diff
@@ -3,23 +3,67 @@ title: TanStack Table Parsers
 description: Store your table state in the URL, with style.
 ---
 
+import { TanStackTableFiltering } from './tanstack-table-filtering.generator'
```
franky47, line 6: **note**: that import doesn't exist and causes the build to fail.
```suggestion
```

acharlop:
@franky47 I've updated the pagination example with 2 different strategies, and if that works for you I'll do the same for the sorting as well.

acharlop:
@franky47 sorry this sat for so long, I completely forgot about it! Rebased on next and finished it up:

- implemented your sorting suggestions (separator, direction format, multi-sort)
- added filtering and a playground demo
- uses TanStack Table v9 (new dep in docs)

FYI: noticed useQueryState keeps the first parser if it's recreated, might be worth an issue?

PR #753, opened 2024-11-09
doc: add break word all for better ux

rasvanjaya21:
Implemented #752

### DIFF
--- a/demo-tabs.tsx
+++ b/demo-tabs.tsx
@@ -24,7 +24,7 @@ export function Demo() {
 				onChange={e => setHello(e.target.value || null)}
 				data-interacted={Boolean(hello)}
 			/>
-			<p>Hello, {hello || 'anonymous visitor'}!</p>
+			<p className="break-all">Hello, {hello || 'anonymous visitor'}!</p>
 		</>
 	)
 }
--- a/packages/docs/content/docs/basic-usage.mdx
+++ b/packages/docs/content/docs/basic-usage.mdx
@@ -28,7 +28,7 @@ export function Demo() {
     <>
       <input value={name || ''} onChange={e => setName(e.target.value)} />
       <button onClick={() => setName(null)}>Clear</button>
-      <p>Hello, {name || 'anonymous visitor'}!</p>
+      <p className="break-all">Hello, {name || 'anonymous visitor'}!</p>
     </>
   )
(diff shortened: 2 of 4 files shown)


packages/docs/content/docs/basic-usage.mdx
```diff
@@ -28,7 +28,7 @@ export function Demo() {
     <>
       <input value={name || ''} onChange={e => setName(e.target.value)} />
       <button onClick={() => setName(null)}>Clear</button>
-      <p>Hello, {name || 'anonymous visitor'}!</p>
+      <p className="break-all">Hello, {name || 'anonymous visitor'}!</p>
```
franky47, line 31: **suggestion**: styles are not needed in documentation code blocks, they distract from the main intent (showing how to use nuqs).


```suggestion
      <p>Hello, {name || 'anonymous visitor'}!</p>
```

demo-tabs.tsx
```diff
@@ -24,7 +24,7 @@ export function Demo() {
 				onChange={e => setHello(e.target.value || null)}
 				data-interacted={Boolean(hello)}
 			/>
-			<p>Hello, {hello || 'anonymous visitor'}!</p>
+			<p className="break-all">Hello, {hello || 'anonymous visitor'}!</p>
```
franky47, line 27: **suggestion**: this file isn't used in the codebase, it was just a demo of how GitHub uses search params to store state, we should probably remove it.


```suggestion
			<p>Hello, {hello || 'anonymous visitor'}!</p>
```

packages/docs/src/app/(pages)/_landing/demo.tsx
```diff
@@ -21,7 +21,7 @@ export async function LandingDemo() {
           <div className="mb-4 h-[136px] animate-pulse rounded bg-zinc-50 dark:bg-zinc-900 sm:h-10" />
         }
       >
-        <div className="mb-4 flex flex-col gap-4 sm:flex-row sm:items-center">
```
franky47, line 24: Looks like this change broke the layout on mobile, the items should be full width:

franky47 (changes requested):
Thanks, the mobile layout needs looking into.

franky47 (approved):

PR #769, opened 2024-11-15
fix(serializer/types): Allow `null` value in serializer for parsers with default values

MartinCura:
See [related discussion](https://github.com/47ng/nuqs/discussions/767).

Passing `null` to the serialize of a search param whose parser has a default currently is a type error, even though the behaviour works as expected.

This is because for the values argument it infers the types using the same `inferParserType` which is fine for parsing but is not quite right for serializing. The way i'm fixing it is just declaring a different `inferSerializerRecordType` to be used here. If there's some TS-foo i don't know which allows reusing the same `inferParserRecordType`, that'd be even better!

`CONTRIBUTING.md` mentions "adding a minimal reproduction in the playground can be very helpful" but i couldn't immediately find how to create one [there](https://nuqs.47ng.com/playground).

Added some very basic tests, feel free to point out if any more would be nice.

### DIFF
--- a/packages/nuqs/src/serializer.ts
+++ b/packages/nuqs/src/serializer.ts
@@ -1,9 +1,18 @@
 import type { Options } from './defs'
-import type { inferSerializerRecordType, ParserBuilder } from './parsers'
+import type { ParserBuilder } from './parsers'
 import { renderQueryString } from './url-encoding'
 
 type Base = string | URLSearchParams | URL
 type ParserWithOptionalDefault<T> = ParserBuilder<T> & { defaultValue?: T }
+
+type inferSingleSerializerType<Parser> =
+  Parser extends ParserBuilder<infer Value> ? Value | null : never
+
+export type inferSerializerRecordType<
+  Map extends Record<string, ParserBuilder<any>>
+> = {
+  [Key in keyof Map]: inferSingleSerializerType<Map[Key]>
+}
 
 export function createSerializer<
   Parsers extends Record<string, ParserWithOptionalDefault<any>>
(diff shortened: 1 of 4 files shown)


packages/nuqs/src/serializer.ts
```diff
@@ -16,7 +25,7 @@ export function createSerializer<
     urlKeys?: Partial<Record<keyof Parsers, string>>
   } = {}
 ) {
-  type Values = Partial<inferParserType<Parsers>>
+  type Values = Partial<inferSerializerRecordType<Parsers>>
```
franky47, line 28: Ah I see, I read your description too quickly, sorry.

**suggestion**: We could still use the inferParserType here, but mark the output as nullable (the import is in `defs.ts`):

```suggestion
  type Values = Partial<Nullable<inferSerializerRecordType<Parsers>>>
```

MartinCura: aah you're quite right, for some reason i thought the type was a bit more complex, and also didn't see you had a `Nullable` type; i'll make the change now

MartinCura: should i switch `inferParserType` for `inferParserRecordType`? the first is still constrained to the latter by the generic of `createSerializer`, but maybe it's a bit more specific and clearer

MartinCura: done it with the original just in case, and to avoid having to change exports in case you didn't want to leak `inferParserRecordType`

franky47 (approved):
LGTM, thanks!

franky47:
Feel free to test the pkg.pr.new build to see if it works on your side.

packages/nuqs/src/serializer.ts
franky47: Yeah we can only expose the one that does everything, let TypeScript deal with the internals. It might be a tiny bit more performant to specialise it, but oh well.

franky47:
The changelog is generated by semantic-release in the GitHub release when the package gets published.

I think we can omit the docs, this is the expected behaviour. Thanks again for your contribution! 🙏

PR #770, opened 2024-11-15
feat: Allow the `serialize` function in parsers to remove their key from the URL

franky47:
Returning `null` from the `serialize` function signals that the associated search param should be removed from the URL.

This would allow clearing invalid states passed to the serialize function.

Edit: this would actually count as a breaking change (the docs did break), so it will be added to the v3 release line.

## Tasks

- [ ] Add docs
- [ ] Add tests

See [related discussion](https://github.com/47ng/nuqs/discussions/662).


franky47:
Closing for now, to be reopened for v3.

PR #788, opened 2024-11-28
Add community adapters section to documentation

Joehoel:
Related to #787. It would be nice to have a section to find the community adapters in the sidebar.

2024-11-28 at 19 36 58@2x

### DIFF
--- a/packages/docs/content/docs/community-adapters/inertia.mdx
+++ b/packages/docs/content/docs/community-adapters/inertia.mdx
@@ -1,0 +1,79 @@
+---
+title: Inertia
+description: Integrate nuqs with Inertia
+---
+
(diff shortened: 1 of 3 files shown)


franky47:
I wonder if we should move the community adapters & parsers into their own section, maybe at some point we could also share community recipes (parser definitions + custom hooks, à la shadcn/ui).

Joehoel:
Maybe a section like this?

```
- Adapters
  - Built in
  - Community
- Parsers
  - ...
```

franky47:
That could work too. I was thinking more like:

```
** Getting started **
- Installation
- Adapters
- Basic usage

** Guides **
...

** Going further **
...

** Community **
- Adapters
  - Inertia
  - One
- Parsers
  - Date & Time
  - Map & Set
  - Compound parsers
  - TanStack Table
```

Issue #206, opened 2021-03-30
Add useHashState


franky47:
`useQueryState` uses the query string part of the URL, which is sent to the server on page requests.
`useHashState` would have the same behaviour (querystring-encoded key/value parameters), but under the `#` part of the URL, which always remains on the client.

This could be useful for end-to-end encrypted apps and libraries like [`47ng/simple-e2ee`](https://github.com/47ng/simple-e2ee).

franky47:
Out of scope for this library (it deals with querystrings, not hashes).

jnachtigall:
Hi @franky47 , thanks for the library which is super helpful. However, we only use it in **client-side rendering** only on Next.js 14 (app router). Now since the app uses query params we ran into the issue, that

1. The URLs get too long (https://nuqs.47ng.com/docs/limits#max-url-lengths) causing [414 URI Too Long](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/414) error not in the browser but rather in our Apache proxy because Apache's [LimitRequestLine](https://httpd.apache.org/docs/current/mod/core.html#limitrequestline) is 8190 by default (8k). Before you ask: Yes, in our app it makes sense to have such a long URL (really :-) ) and modern browsers have no problem with it. Our app can grow up to about 20k length.
2. Using query params instead hash `#` fragments does result in hardly URLs that are hardly possible to **cache** in caching proxies like Varnish or Squid.

Since we only use `nuqs` client side but not for SSR, technically it would be wiser to use hash fragments anyway. Maybe you could reconsider your choice and add this add an option.

For the time being, unfortunately, I think we'll have to remove `nuqs` and replace it with something like https://github.com/hejmsdz/use-hash-param which unfortunately is far more limited in options (parsing for instance). Maybe you also know another alternative to use if one wants hashes and not query params? Naming those in the docs might also be possible, because there definitely is a use case.

franky47:
This might actually be possible now with adapters, albeit not out-of-the-box.

There are still parts of the core codebase that reads from `location.search`, that would have to be abstracted away in the adapter interface, for it to use other storage mechanisms (hash/fragment indeed, but localStorage could also be an candidate).

This might also be needed if we want to target React Native / Expo Router at some point, which might not be based on a Location object (TBC).

jnachtigall:
[...]

PR #800, opened 2024-12-10
feat: Abstract reading from location.search in adapters interface

franky47:
This is mainly for the testing adapter to only work on the provided set of initial search params without needing to stub/mock location.search.

But it also opens the door for adapters to define new storage locations, like hash/fragment or even localStorage.


franky47:
Note: might be useful for #723 and #206.

PR #818, opened 2024-12-19
feat: Add loader feature 

franky47:
While we have the cache feature in Next.js app router to parse the incoming `searchParams` page prop and provide it to the rest of the RSC tree (via React's `cache` function, which is a hack but not the subject of this PR), other frameworks lacked the ability to parse an incoming request's search params and get the same kind of type-safe object back (without the caching part).

This PR introduces the `createLoader` function, a factory that takes in a search params descriptor object (a "parsers map" as we call it internally), and returns a `load` function. That function can then process a lot of different kinds of inputs:
- URL object
- Request object
- URLSearchParams object
- A string, whether of a fully qualified URL, or just the `search` part
- A record of key/values as what Next.js usually gives us (`Record<string, string | string[] | undefined>`)

Example usage in Remix / React Router's loaders:

```ts
import { createLoader, parseAsInteger } from 'nuqs/server' // can also be imported from 'nuqs' in non-RSC frameworks

const searchParams = {
  pageIndex: parseAsInteger.withDefault(1),
  pageSize: parseAsInteger.withDefault(10)
}

const loadSearchParams = createLoader(searchParams, { 
  urlKeys: {
    pageIndex: 'page',
    pageSize: 'limit'
  }
})

// When called from /path?page=2&limit=50
export function loader({ request }: LoaderFunctionArgs) {
  const { pageIndex, pageSize } = loadSearchParams(request)
  // ^? { pageIndex: 2, pageSize: 50 }
  return fetchData(pageIndex, pageSize) 
}
```

## Compatibility

- [x] React Routers & Remix loaders (using a `Request` object)
- [x] Next.js pages router `getServerSideProps` (needs a test)
- [x] Next.js app router `searchParams` prop (async in Next 15 -> good for PPR)
- [x] Next.js API routes

## Tasks

- [x] Add docs


packages/docs/content/docs/server-side.mdx
```diff
@@ -3,6 +3,126 @@ title: Server-Side usage
...
+<Callout type="warn" title="Note">
+  Loaders **don't validate** your data. If you expect positive integers
+  or JSON-encoded objects of a particular shape, you'll need to feed the result
+  of the loader to a schema validation library, like [Zod](https://zod.dev).
+
+  Built-in validation support is coming. [Read the RFC](https://github.com/47ng/nuqs/discussions/446).
+  Alternatively, you can build validation into [custom parsers](/docs/parsers/making-your-own).
+</Callout>
+
+
```
franky47, line 115: Somehow a line got lost in the edits here, it should read:

```suggestion
The loader function will accept the following input types to parse search params from:
```

packages/docs/content/docs/server-side.mdx
```diff
@@ -3,6 +3,126 @@ title: Server-Side usage
...
+export const coordinatesSearchParams = {
+  latitude: parseAsFloat.withDefault(0)
+  longitude: parseAsFloat.withDefault(0)
+}
+
+export const loadSearchParams = createLoader(coordinatesSearchParams)
+```
+
+Here, `loadSearchParams{:ts}` is a function that parses search params and returns
+state variables to be consumed server-side (the same state type that `useQueryStates{:ts}` returns).
```
franky47, line 27:
```suggestion
state variables to be consumed server-side (the same state type that [`useQueryStates{:ts}`](/docs/batching) returns).
```

packages/docs/content/docs/server-side.mdx
```diff
@@ -3,6 +3,126 @@ title: Server-Side usage
...
+
+export default async function Page({ searchParams }: PageProps) {
+  const { latitude, longitude } = await loadSearchParams(searchParams)
+  return <Map
+    lat={latitude}
+    lng={longitude}
+  />
+
+  // Pro tip: you don't *have* to await the result.
+  // pass the Promise object to children components wrapped in Suspense
```
franky47, line 48:
```suggestion
  // Pass the Promise object to children components wrapped in <Suspense>
```

packages/docs/content/docs/server-side.mdx
```diff
@@ -3,6 +3,126 @@ title: Server-Side usage
...
+  Alternatively, you can build validation into [custom parsers](/docs/parsers/making-your-own).
+</Callout>
+
+
+- A string containing a fully qualified URL (eg: `https://example.com/?foo=bar`)
+- A string containing just search params (like `location.search`, eg: `?foo=bar`)
+- A `URL{:ts}` object
+- A `URLSearchParams{:ts}` object
+- A `Request{:ts}` object
+- A `Record<string, string | string[] | undefined>{:ts}` (eg: `{ foo: 'bar' }{:ts}`)
```
franky47, line 121:
```suggestion
- A string containing a fully qualified URL: `https://example.com/?foo=bar`
- A string containing just search params: `?foo=bar` (like `location.search{:ts}`)
- A `URL{:ts}` object
- A `URLSearchParams{:ts}` object
- A `Request{:ts}` object
- A `Record<string, string | string[] | undefined>{:ts}` (eg: `{ foo: 'bar' }{:ts}`)
```

PR #825, opened 2024-12-26
doc: Fix link to options docs

blyme:
The options is located at https://nuqs.47ng.com/docs/options and not in https://nuqs.47ng.com/docs/parsers/options :)


franky47 (approved):
Ah, good catch, thanks!

PR #830, opened 2024-12-29
chore: Test a theory for RR sync

franky47:
There may be some extra renders being caused by two hooks emitting the same sync event due to parallel effects running.

Edit: coding on a phone is not the best DX..


franky47:
I found that patching history is needed to follow Links in React Router frameworks. So it's going to be enabled by default (and not exported). I'll close this PR and cherry pick the relevant parts in a more focused, better documented PR.

PR #847, opened 2025-01-05
doc: Update wording for the "missing suspense boundary" guidelines

mtlaso:
Updating the wording of [Client components need to be wrapped in a <Suspense> boundary](https://nuqs.47ng.com/docs/troubleshooting#client-components-need-to-be-wrapped-in-a-suspense-boundary) section based on the discussion at https://github.com/47ng/nuqs/discussions/846


franky47 (approved):

franky47:
Thanks!

PR #848, opened 2025-01-06
doc: Add Waku community adapter

rmarscher:


rmarscher:
Backlink - https://github.com/dai-shi/waku/discussions/1123

It works with SSR if the Waku parent component render setting is "dynamic".

I want to see what we can do on the server / RSC side of things. I think this is only a client and SSR implementation so far.

rmarscher:
Ha. It totally works with RSC! I updated the adapter to support the scroll and shallow options. 

Fun demo: https://waku-nuqs.martiandev.workers.dev/

packages/docs/content/docs/community-adapters/waku.mdx
```diff
@@ -0,0 +1,90 @@
...
+"use client";
+
+import {
+  type unstable_AdapterOptions as AdapterOptions,
+  unstable_createAdapterProvider as createAdapterProvider,
+  renderQueryString,
+} from "nuqs/adapters/custom";
+import { useRouter_UNSTABLE as useRouter } from "waku";
+
+export function renderURL(base: string, search: URLSearchParams) {
```
franky47, line 25: **question**: Does the renderURL function need to be exported for other parts of Waku to leverage?

packages/docs/content/docs/community-adapters/waku.mdx
```diff
@@ -0,0 +1,90 @@
...
+  const hashlessBase = base.split("#")[0] ?? "";
```
franky47, line 26: **note**: Based on usage here, the `base` parameter would never contain a hash part (it's fed `location.origin + location.pathname`).

franky47 (approved):

packages/docs/content/docs/community-adapters/waku.mdx
rmarscher: Ah yeah, thanks. I simplified that. Waku's useRouter returns the current path, so I also added using that. We can preserve the hash when it goes through RSC. I updated the adapter to do that. I think including `location.origin` isn't necessary so I dropped it.

packages/docs/content/docs/community-adapters/waku.mdx
```diff
@@ -22,25 +22,21 @@ import {
...
-  return hashlessBase + query + hash;
-}
-
 function useNuqsAdapter() {
-  const { query, push, replace } = useRouter();
+  const { path, query, push, replace } = useRouter();
   const searchParams = new URLSearchParams(query);
   const updateUrl = (search: URLSearchParams, options: AdapterOptions) => {
+    const query = renderQueryString(search);
+    const hash = typeof location !== "undefined" ? location.hash : "";
```
franky47, line 30: Does this mean this can be called on the server somehow?

useQueryState(s) are client-side hooks, that can render on the server in an SSR pass, but the state updater function should only be called client-side.

rmarscher: OK, great. I assumed that was probably case, but the guard seemed to be safe just in case it was somehow run in an environment without a location global. I'll take it out.

rmarscher: Done.

PR #855, opened 2025-01-09
feat: Key isolation in the React SPA adapter

franky47:
Key isolation is the fact that updating a search param only re-renders hooks subscribed to this key.

Most frameworks (Next.js & React Router) re-render every call site of their useSearchParams hooks when any search params change, causing unnecessary re-renders.

Since in the React SPA we control the whole thing (there is no router so to speak), we can implement caching to detect only relevant differences and avoid re-rendering unnecessarily.


packages/e2e/shared/specs/render-count.cy.ts
```diff
@@ -63,9 +50,9 @@ export function testRenderCount({
           if (props.delay) {
             cy.wait(props.delay)
           }
-          assertLogCount('render', expected.mount + expected.update)
           cy.get('#state').should('have.text', 'pass')
           cy.location('search').should('contain', 'test=pass')
+          assertLogCount('render', expected.mount + expected.update)
```
franky47, line 55: note: Unrelated change to help the render count be less flaky: by asserting the log count after testing for correctness, Cypress will wait for the URL to be correct (and give logs time to flush) before testing.

packages/nuqs/src/adapters/react.ts
```diff
@@ -12,30 +17,38 @@ function updateUrl(search: URLSearchParams, options: AdapterOptions) {
   url.search = renderQueryString(search)
   const method =
     options.history === 'push' ? history.pushState : history.replaceState
-  method.call(history, history.state, '', url)
+  method.call(history, history.state, historyUpdateMarker, url)
```
franky47, line 20: **note**: Looks like it this should be a bug fix for external history updates & sync optimisations, but it doesn't seem to make a difference. Better be aligned with the patch-history expected behaviour anyway.

packages/nuqs/src/useQueryStates.ts
```diff
@@ -83,7 +83,7 @@ export function useQueryStates<KeyMap extends UseQueryStatesKeysMap>(
       ),
     [stateKeys, urlKeys]
```
franky47, line 84: **note**: urlKeys isn't guaranteed to be stable (if defined inline, it's not), so this would cause a cascade of referential instability.

Add a test & fix:
```suggestion
    [stateKeys, JSON.stringify(urlKeys)]
```

michaelbayday:
@franky47 are there any solutions you can think of that we can implement currently that would prevent excessive re-renders? we're currently use deckgl to lift state from the url but its causing multiple rerenders which results in unexpected layer rendering

franky47:
The React Compiler is one that comes to mind, it would memoise components so they don't rerender if their props are the same. It may require you to slice your components a little differently, with a parent component doing the URL state access, and feeding that (or derived props) to heavy children components that'd be target for auto-memoisation.

You could also do the same thing the compiler does, with React.memo, but it's a lot more work and can be brittle, more here:

http://podrocket.logrocket.com/the-useless-usecallback-react-performance-myths-dominik-dorfmeister

PR #900, opened 2025-02-07
feat: Debounce

### LINKED ISSUE #291: Debounce URL updates
Hello,
For now, it cannot be used to save the state of a search input because the URL is updated for each key stroke. It would be nice to debounce or throttle the URL updates for a smoother interaction ;)

franky47:
## Try it

```
pnpm add https://pkg.pr.new/nuqs@900
```

## Documentation

This PR introduces a new method of rate-limiting URL updates, via debouncing rather than throttling (which was already possible via the `throttleMs` option).

### Debouncing vs throttling

https://kettanaito.com/blog/debounce-vs-throttle

Throttling gives you _predictable, consistent_ updates at a given rate. It's necessary to mitigate rate limits of the browser History API methods. One advantage of throttling is that, assuming there was sufficient inactivity before an event occurred, it will be emitted instantly, allowing for a reactive first update (and delaying subsequent ones).

Debouncing on the other hand gives you _one eventual_ update after a given amount of inactivity has passed. Debouncing is useful for high-frequency inputs where only the final value is valuable, like moving a slider or typing some text.

Both of these mechanisms are in place to rate-limit _updates to the URL_. The internal state returned by `useQueryState(s)` is always updated instantly (just like `React.useState` would).

### Specifying options

The `throttleMs` option is now marked as deprecated (but still supported until a later major, maybe v4, TBC), and replaced with a `limitUrlUpdates` option:

```ts
const [search, setSearch] = useQueryState('q', {
  limitUrlUpdates: {
    method: 'debounce', // debounce | throttle
    timeMs: 500
  }
})

// Can be defined with the builder pattern:
parseAsString.withOptions({ limitUrlUpdates: {...} })

// Can also be passed as a call option:
setSearch('foo', { limitUrlUpdates: {...} })
```

#### Shorthands

You can use the `throttle` and `debounce` shorthand methods, as well as the `defaultRateLimit` export to shorten your options:

```ts
import { debounce, throttle, defaultRateLimit } from 'nuqs' // or 'nuqs/server' in Next.js app router server code

parseAsString.withOptions({ limitUrlUpdates: debounce(500) })
parseAsString.withOptions({ limitUrlUpdates: throttle(500) })
setSearch(e.target.value, { limitUrlUpdates: defaultRateLimit }) // Reset to default
```

### Interaction with other options

The `startTransition` function is used only when the URL is being updated, so for a debounced update, the loading state will only be triggered once the debounce time has elapsed, not while it is pending.

You'll likely want to set `shallow: false` to notify the server of updates, as I don't see many reasons to debounce a client-only URL update.

If you set both `throttleMs` and `limitUrlUpdates`, `limitUrlUpdates` will take precedence.

### Tips & tricks

You will likely want to turn on debounce on high-frequency state updates. But often, inputs have multiple source of updates with different event firing rates, for example:
- A slider (`<input type="range">`) has a very high event emitting frequency for `onChange` on drag that you may want to debounce, but clicking on a point on the slider track (not dragging) should likely be an instant update.
- A text input has a high event emitting frequency for `onChange` when typing that warrant debouncing, but actions like clearing the input (select all + backspace) or pressing Enter should commit the value instantly.

One recommendation is to define debouncing when calling the state updater function:

```tsx
function Search() {
  const [search, setSearch] = useQueryState("q")
  return (
    <input
      value={search}
      onChange={(e) =>
        setSearch(e.target.value, {
          limitUrlUpdates: e.target.value === '' ? undefined : debounce(500),
        })
      }
      onKeyDown={(e) => {
        if (e.key === "Enter") {
          // Send the search immediately when pressing Enter
          setSearch(e.currentTarget.value)
        }
      }}
    />
  )
}
```

## Internal

Prep work:

- Move shared code into src/lib, leaving only features at the top level
- Move the throttle queue into a class to help with testing &
  multi-storage support later
- Add stubs for debounce queue
- Implemented new option
- Added debounce queues (one per key) in front of the global throttle queue

## Tasks
- [x] Tests
- [x] Documentation
- [x] Memory usage: cleanup debounce queues after usage, cleanup timeout abort event listeners
- [x] Handle infinity timeouts
- [x] Add abort method to cancel a pending debounced update 
- [x] Make auto-abort on navigation work in the Next.js app router
- [x] Elect repro-702 to shared e2e test (to validate queued queries)
- [x] Restore the default behaviour of adapters (auto-resetting the throttled queue) to avoid breaking custom ones.
- ~Investigate double-update mitigation (debounced update followed by a pre-empted default set, eg: pressing Enter to submit immediately on a debounced input).~

See discussions #373 & #449.

Closes #291.


Kavan72:
Can't wait to test this feature 😁

franky47:
@Kavan72 you can! Follow rhe preview install instructions here, and the docs from the PR description:
> ```
> npm i https://pkg.pr.new/nuqs@900
> ```

Kavan72:
any update on this ? 🙂

franky47:
Have you had a chance to test it? Do you have any feedback?

One issue I'm encountering is that it's possible (well it was with throttle too but harder to trigger) to schedule a search params update, perform a navigation (eg: clicking a Link), and the search params apply on the destination page (potentially overriding some that were applied by the link).

The tests that confirm this behaviour pass on my machine but fail in CI, so I'll have to look into that.

The other point is about cascading updates to the same key with different options. If scheduling a debounced update, then a throttled update, the throttled one will fire first, and the debounced one will overwrite it later. This feels like it breaks the intent of "last update wins".
One case where this is visible is in search fields: you want to debounce in onChange, but want to flush/emit the update immediately in some cases (clearing the search field, or pressing Enter). Those immediate updates should abort the debounced update.

Kavan72:
> Have you had a chance to test it? Do you have any feedback?
> 
> One issue I'm encountering is that it's possible (well it was with throttle too but harder to trigger) to schedule a search params update, perform a navigation (eg: clicking a Link), and the search params apply on the destination page (potentially overriding some that were applied by the link).
> 
> The tests that confirm this behaviour pass on my machine but fail in CI, so I'll have to look into that.
> 
> The other point is about cascading updates to the same key with different options. If scheduling a debounced update, then a throttled update, the throttled one will fire first, and the debounced one will overwrite it later. This feels like it breaks the intent of "last update wins". One case where this is visible is in search fields: you want to debounce in onChange, but want to flush/emit the update immediately in some cases (clearing the search field, or pressing Enter). Those immediate updates should abort the debounced update.

Sorry, I didn’t get a chance to test this, but here are some of my thoughts related to the current issues:
1. If changes are pending and the user navigates away, we have two options: one is to push the changes into history before navigating, so the back button doesn’t break; the other is to simply abort the pending changes and navigate.
2. If an update is pending and the user presses Enter to forcefully update the same key, then, in my opinion, only the last update (the forced one) should be considered.

franky47:
Thanks for your feedback. I agree with point 2, I updated the implementation to do just that yesterday.

For point 1, I didn't think of the "apply before navigating", which makes sense, unfortunately I doubt we'll be able to reliably inject custom behaviours **before** navigation occurs in all frameworks (and even doing so **after** navigation would have its issues).

I was thinking of providing a way to abort pending debounced updates, what do you think?
https://bsky.app/profile/francoisbest.com/post/3lircw7dsuk2d

Willem-Jaap:
This is so cool, going to try this in a project soon!
Mad respect for the amount of thought that seems to have gone into the API design and implementation of this! 🙌 🔥

franky47:
@Willem-Jaap thanks, this will land in beta next week, so any feedback is much appreciated while there is time to steer the API and behaviours before it crystallises in GA.

Elya29:
This perfectly fits my needs for debouncing slider values! Thank you so much for the implementation. I tried the new feature and it’s working perfectly on my side. Can’t wait for the beta to be released!

franky47:
The last hurdle I have with this feature is cancelling pending debounced updates when the user navigates using a Link or an imperative router call. It might only happen on longer debounce timeouts, but it will feel like a bug to have search params being applied on a different pathname and/or erasing a value set by a navigation, what do y'all think about that?

Willem-Jaap:
Yeah, makes sense. I’d cancel any pending debounced updates on route change. Otherwise it feels buggy if a search param suddenly applies on a new pathname or overwrites navigation state. Safe default, predictable behavior in my opinion

Willem-Jaap:
Just tested this in my project, works great!

devhasson:
I can't wait for this PR to be merged 👀

franky47:
@devhasson have you tried it?

It's in a state where there are a lot of internal changes that should not (theoretically) affect userland behaviours, but I'm looking for folks to try it and give feedback of things that might have been missed by the test suite.

Earlier this week I've been working at getting the "auto-abort pending updates on user navigation" detection working, it's now green in all adapters, but I'd like to try it in a couple of actual apps before it lands, just in case we end up with another aha moment like last week.

Willem-Jaap:
Do you have a list of scenario's of which you'd like feedback?

I feel like most of my usecases are pretty standard with regards to status/filter state 

image

franky47:
@Willem-Jaap what framework are you using?

The PR itself has value in any setup, as there are some internal changes when using the default options (throttle set to the default rate), due to trying to cancel those queued updates on navigation.

The scenario would be that of a stress-test one: click around and navigate as fast as you can, and see if things stay correct.

Thanks! 🙏

didemkkaslan:
please merge it :D <3

Willem-Jaap:
^^ I'm using Next.js with Radix primitives for tabs & other stuff

Haven't encountered any issue with selecting as fast as possible, everything seems to keep in sync nicely

jaberamin9:
Please merge this PR. This feature is urgently needed.

PR #902, opened 2025-02-07
doc: add tanstack multiple paginations example

stychu:
Update the documentation with a pagination example how to use multiple pagination/tables with nuqs and search params.

### DIFF
--- a/packages/docs/content/docs/parsers/community/tanstack-table-multiple.generator.tsx
+++ b/packages/docs/content/docs/parsers/community/tanstack-table-multiple.generator.tsx
@@ -1,0 +1,274 @@
+'use client'
+
+import { CodeBlock } from '@/src/components/code-block.client'
+import { Querystring } from '@/src/components/querystring'
+import { Label } from '@/src/components/ui/label'
+import {
+  Pagination,
+  PaginationButton,
+  PaginationContent,
+  PaginationItem,
+  PaginationNext,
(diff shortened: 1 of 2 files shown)


franky47:
Thanks! There are a few other PRs related to TanStack Table parsers & docs, I'll have to see how to best combine them.

Regarding "namespacing" keys in the URL to support multiple instances of a component without state overlap, I was thinking of adding a prefix option for the urlKeys feature, that would still return the state names as declared, but add that prefix automatically in the URL, what do you think?

stychu:
> Regarding "namespacing" keys in the URL to support multiple instances of a component without state overlap, I was >thinking of adding a prefix option for the urlKeys feature, that would still return the state names as declared, but add that >prefix automatically in the URL, what do you think?

Sure that will also do !
 But for our use case if we go with two separate params per pagination the url will become a mess quickly as we have 3-10 tables per url so with array param its much cleaner.

junwen-k:
I am the author for the related PR https://github.com/47ng/nuqs/pull/836 :) Feel free to let me know if there's anything that I can help with adding Tanstack Table as community examples. @franky47

franky47:
@junwen-k thanks, I've had your PR in my review list for way too long (sorry about that), I need to take some time at some point to do a good pass on all docs related PRs.

josealfalf (dismissed):
000*)

> ###

PR #908, opened 2025-02-12
fix: useQueryStates does not update values correctly when config changes

### LINKED ISSUE #907: "useQueryStates" does not update values correctly when config changes
#### What's your version of `nuqs`?

2.3.2 (see the reproduction on CodeSandbox)

#### What framework are you using?

I’ve reproduced the bug in Next.js using both the App router and Pages router.

- ✅ Next.js (app router)
- ✅ Next.js (pages router)

#### Which version of your framework are you using?

- app router: `14.2.21`
- pages router: `15.1.7`

## Description

I'm using the useQueryStates hook to manage page filters in my app. Users can manually control filter visibility by adding or removing filters. Since the useQueryStates configuration is dynamically generated based on the filters received from the parent component, I expect the hook to always return values that match the current configuration.

For example, I initialize a component with the following configuration:

```ts
const [nuqsConfig, setNuqsConfig] = useState({
  p1: parseAsString,
  p2: parseAsString
})

const [values, setValues] = useQueryStates(nuqsConfig)

// values: { p1: null, p2: null }
```

Then, in response to a user action, I update the configuration by removing one of the keys:

```ts
setNuqsConfig({ p1: parseAsString })
```

The component re-renders and passes this new configuration to the "useQueryStates" hook. I expect the hook to update accordingly and return `values === { p1: null }`. However, the "values" variable still contains keys from the previous configuration.

The same issue occurs when adding or replacing keys.

## Reproduction

Code: https://codesandbox.io/p/devbox/ds7nyy

Direct preview link: https://ds7nyy-3000.csb.app/

1. The useQueryStates hook is initialized with the following config: `{ p1: parseAsString, p2: parseAsString }` and returns values in this format: `{ p1: null, p2: null }`
2. Click the "Update config (remove one of the keys)" button
3. The configuration updates to `{ p1: parseAsString }`, but the values object still contains the previous structure: `{ p1: null, p2: null }`, instead of the expected `{ p1: null }`.

maslianok:
Fixes #907

### DIFF
--- a/packages/docs/src/app/playground/_demos/repro-907/page.tsx
+++ b/packages/docs/src/app/playground/_demos/repro-907/page.tsx
@@ -1,0 +1,55 @@
+// https://github.com/47ng/nuqs/issues/907
+
+'use client'
+
+import { parseAsString, useQueryStates } from 'nuqs'
+import { useState } from 'react'
+
+export default function Home() {
+  const [nuqsConfig, setNuqsConfig] = useState({
+    p1: parseAsString,
+    p2: parseAsString
+  })
+
+  const [values] = useQueryStates(nuqsConfig)
+
+  return (
+    <>
+      <div className="flex gap-2">
+        <button
+          onClick={() => setNuqsConfig({ p1: parseAsString })}
+          className="border p-2"
+        >
+          Update config (remove one of the keys)
+        </button>
+
+        <button
+          onClick={() =>
+            setNuqsConfig({
+              p1: parseAsString,
+              p2: parseAsString,
+              p3: parseAsString
+            })
+          }
+          className="border p-2"
+        >
+          Update config (add a new key)
+        </button>
+
+        <button
+          onClick={() =>
+            setNuqsConfig({
+              p1: parseAsString,
+              p5: parseAsString
+            })
+          }
+          className="border p-2"
+        >
+          Update config (replace a key)
+        </button>
+      </div>
+      <div>Config keys: {JSON.stringify(Object.keys(nuqsConfig))}</div>
+      <div>Result: {JSON.stringify(values)}</div>
+    </>
+  )
+}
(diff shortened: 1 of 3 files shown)


packages/docs/src/app/playground/_demos/repro-907/page.tsx
```diff
@@ -0,0 +1,55 @@
...
+'use client'
+
+import { parseAsString, useQueryStates } from 'nuqs'
+import { useState } from 'react'
+
+export default function Home() {
+  const [nuqsConfig, setNuqsConfig] = useState({
+    p1: parseAsString,
+    p2: parseAsString
+  })
```
franky47, line 12: This fails type-checking due to the {p1, p2} object being narrowed down to that specific shape.

**suggestion**: Widen the type to accept any key
```suggestion
  } as Record<string, ParserBuilder<string>>)
```

packages/nuqs/src/useQueryStates.ts
```diff
@@ -310,6 +310,16 @@ function parseMap<KeyMap extends UseQueryStatesKeysMap>(
...
+
+  if (!hasChanged) {
+    // check that keyMap keys have not changed
+    const keyMapKeys = Object.keys(keyMap)
+    const cachedStateKeys = Object.keys(cachedState ?? {})
+    hasChanged =
+      keyMapKeys.length !== cachedStateKeys.length ||
+      keyMapKeys.some(key => !cachedStateKeys.includes(key))
+  }
+
```
franky47, line 322: **note**: This check was already made on line 115-116, so if we get into that if block we should disregard the `hasChanged` flag: we know the structure has changed so we should always set the internal state and stateRef.

**suggestion**: Remove this block here in parseMap, and the `if (hasChanged)` check on line 125, to always update the state.

franky47 (changes requested):
Thanks for the repro and the test!

~The fix can be made simpler in `useQueryStates.ts`, if we get into the block where we detected the keys changed (lines 114-135), we can safely disregard the hasChanged flag and reset the state.~

franky47 (approved):

PR #909, opened 2025-02-12
fix: Handle console.log of null values in React Devtools

### LINKED ISSUE #808: [Chrome Extension] console debug mode will crash crash your app when using localStorage.setItem("debug", "nuqs")
## Context

```
    "nuqs": "^2.2.3",
```

What framework are you using?

- ✅ Next.js (app router)

14.2.20

## Description & reproduction
On localhost; go to console and do:
```
localStorage.setItem("debug", "nuqs")
```
Then go to any page with parameters that have default value. For me it was a boolean parsen "open" with default false. Then if you go to any page it will give "cannot parse null to string";

image

franky47:
In some cases (can't reproduce it reliably though), turning on debug logs and logging console.log('%s', null) will throw an error due to this line in React Devtools: 

https://github.com/facebook/react/blob/f83903bfcc5a61811bd1b69b14f0ebbac4754462/packages/react-devtools-shared/src/backend/utils/formatConsoleArguments.js#L59-L61

PR fixing it: https://github.com/facebook/react/pull/32372.

Falling back to the flattened message if the overloaded console.log fails.

Closes #808.

### DIFF
--- a/packages/nuqs/src/debug.test.ts
+++ b/packages/nuqs/src/debug.test.ts
@@ -1,8 +1,8 @@
-import { describe, expect, test } from 'vitest'
+import { describe, expect, it } from 'vitest'
 import { sprintf } from './debug'
 
 describe('debug/sprintf', () => {
-  test('%s', () => {
+  it('formats strings with %s', () => {
     expect(sprintf('%s', 'foo')).toBe('foo')
     expect(sprintf('%s', 1)).toBe('1')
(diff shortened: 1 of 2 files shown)


franky47 (commented):
#808 being closed by #909 is the drum machine reference I didn't expect today 🤘🥁

franky47 (commented):
Urgh never comment in the VSCode PR view, always creates double comments 😓

PR #805, opened 2024-12-12
feat: Add temporary dedicated server features imports

franky47:
This allows using the serializer and the parsers (and later on the loaders) in server-side code without depending on React for the cache, as would be the case if importing from `nuqs/server`.

🦋 https://bsky.app/profile/francoisbest.com/post/3ld5bqxov722z
X: https://x.com/nuqs47ng/status/1867331081349083400

## Tasks

- [x] Update docs
- [x] Move everything under a single path named explicitly (`nuqs/server/temporary-react-agnostic` or something)
- [x] Rebase and include loaders
- [ ] Investigate import issues with `moduleResolution` -> try moving the export to `nuqs/server-temporary-react-agnostic` if it's a nesting issue.

## Note

This PR adds three new imports:

- `nuqs/server/cache`: exports only the cache feature. Requires React canary (currently only useful in Next.js app router)
- `nuqs/server/parsers`: exports defs & all built-in parsers
- `nuqs/server/serializer`: exports defs & the `createSerializer` feature

The `nuqs/server` import is untouched in this PR.

## Future breaking change plans

In nuqs v3:
- `nuqs/server/cache` will be the only way to import the cache feature, it will be removed from `nuqs/server`
- `nuqs/server/parsers` and `nuqs/server/serializers` may be removed for cleanup, as their code will be importable from `nuqs/server` in a framework-agnostic way.

Closes #804.
See also discussion #717.

PR #923, opened 2025-02-21
fix: Import React.cache only when needed

### LINKED ISSUE #804: `nuqs/server` requires `react`
## Context

```
"nuqs": "^2.2.3",
```

What framework are you using?

- ✅ Next.js (pages router)

```
  next: 14.2.13 // An outdated version detected (latest is 15.1.0), upgrade is highly recommended!
  eslint-config-next: N/A
  react: 18.3.1
  react-dom: 18.3.1
  typescript: 5.6.3
```

## Description

Apart from Frontend usage, I wanted to use my nuqs parsers to generate URLs on the server-side to be sent in emails.

For that I am trying to use a serializer on an API Route like so:

```typescript
import {
  createSerializer,
  parseAsStringLiteral,
} from "nuqs/server";

export const universalQueryParams = {
  "user-menu": parseAsStringLiteral(["open", "closed"]).withDefault("closed"),
};

const serializer = createSerializer(universalQueryParams);

export const serializeUniversalQueryParams = serializer(base, values);
```

But sadly my Next.js-build fails, telling me:

```logs
[build:next] Failed to compile.
[build:next] 
[build:next] ./node_modules/nuqs/dist/server.js
[build:next] Attempted import error: 'cache' is not exported from 'react' (imported as 'cache').
[build:next] 
[build:next] Import trace for requested module:
[build:next] ./node_modules/nuqs/dist/server.js
```

Is there a trick to it? As I see it, this module is required for the cache feature, that I don't need. Would splitting up the path to allow direct imports of parsers/serializer help?

franky47:
Having `import { cache } from 'react'` caused issues when using React 18 or 19 GA, because the `cache` function is only exported in canary builds (which the Next.js app router uses internally, regardless of what's in your app's package.json).

This meant importing from `'nuqs/server'` caused an import error when done from non app-router code, like the pages router, an API route definition, or a non-Next.js framework, which would fallback to the version of React defined in the package.json (likely a stable one).

Changing the import to `import * as React from 'react'` is what is being highlighted in the React docs themselves, and allows to only call the cache function when actually creating a cache object.

Closes #804, and supersedes #805.

### DIFF
--- a/packages/nuqs/src/cache.ts
+++ b/packages/nuqs/src/cache.ts
@@ -1,5 +1,4 @@
-// @ts-ignore
-import { cache } from 'react'
+import * as React from 'react'
 import type { SearchParams, UrlKeys } from './defs'
 import { error } from './errors'
 import { createLoader } from './loader'
@@ -27,7 +26,7 @@ export function createSearchParamsCache<Parsers extends ParserMap>(
   // whereas a simple object would be bound to the lifecycle of the process,
   // which may be reused between requests in a serverless environment
   // (warm lambdas on Vercel or AWS).
-  const getCache = cache<() => Cache>(() => ({
+  const getCache = React.cache<() => Cache>(() => ({
     searchParams: {}
   }))
 


rik-iso:
```
web:build | ../../node_modules/.pnpm/nuqs@2.4.1_next@14.2.23_@babel+core@7.24.6_@opentelemetry+api@1.9.0_react-dom@18.3.1_react@18_254swa4g7i3oqurqrbun7hl3xa/node_modules/nuqs/dist/server.js
web:build | Attempted import error: 'cache' is not exported from 'react' (imported as 'React').

web:build | Import trace for requested module:
web:build | ../../node_modules/.pnpm/nuqs@2.4.1_next@14.2.23_@babel+core@7.24.6_@opentelemetry+api@1.9.0_react-dom@18.3.1_react@18_254swa4g7i3oqurqrbun7hl3xa/node_modules/nuqs/dist/server.js
web:build | ./app/features/search/(sections)/products/hooks/filters/getProductSearchFilterParsers.ts
web:build | ./app/features/search/(sections)/products/hooks/filters/useProductSearchQueryStates.ts
web:build | ./components/page/marketplace/MarketplaceProducts.tsx
```

still seeing this issue on 2.4.1

franky47:
@rik-iso what's your setup?

rik-iso:
This might be operator error but I'm confused how to fix it.

The specific file `getProductSearchFilterParsers` is importing:
```
import { parseAsArrayOf, parseAsInteger, parseAsString, parseAsStringEnum, parseAsStringLiteral } from 'nuqs/server'
```

In the above crash, it's being used in a pages router page, but it also needs to be used in an app router page (we're mid-migration). Should I be importing it differently?

rik-iso:
(The crash actually only happens on  `next build` and not in my dev server)

franky47:
Are you using the cache feature (`createSearchParamsCache`) anywhere at all?

I'll try and reach out to the Next.js team to see if they have ideas for this kind of use-case, and I'll try and reproduce on my end in the mean time.

rik-iso:
Yeah I am using that cache on this property but only in the app router pages - the cache shouldn't be being used in the pages router code path.

rik-iso:
I'm going to see if I can refactor out the use of the cache as I think it might be mostly redundant and see if that stops the error for now.

franky47:
If you don't need the no-prop-drilling property of the cache, you can use a [loader](https://nuqs.47ng.com/docs/server-side#loaders) instead.

rik-iso:
Possibly found the culprit - for some reason I am doing 

```
export type SearchQueryState = ReturnType<typeof searchQueryStateCache.all>
export const searchQueryStateCache = createSearchParamsCache(getProductSearchFilterParsers(SEARCH_DEFAULT_PARSERS))
```

then I am only importing `SearchQueryState` using `import type` into a pages router page... but maybe that's enough to break the build (but not dev server).

Cleaning this up to use `inferParserType`... it's code I wrote in early nuqs 1 days so maybe that didn't exist then or I didn't know about it 🙃 .

rik-iso:
nevermind, that didn't solve it sadly.

franky47:
Could you paste the result from running `next info` please?

While trying to reproduce this in the e2e environment, I managed to get the cache to compile in the pages router, which is definitely not supposed to happen 🙃 (ie: I can't get it to fail on `next build`).

rik-iso:
```
 pnpm next info
Detected shared packages: @packages/icons, @packages/image-optimization, @packages/react-icons, @packages/react-utils, @packages/utils, @packages/dom-utils

Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.3.0: Thu Jan  2 20:24:23 PST 2025; root:xnu-11215.81.4~3/RELEASE_ARM64_T6020
  Available memory (MB): 32768
  Available CPU cores: 12
Binaries:
  Node: 22.13.1
  npm: 9.6.7
  Yarn: N/A
  pnpm: 9.15.4
Relevant Packages:
  next: 14.2.23 // An outdated version detected (latest is 15.2.1), upgrade is highly recommended!
  eslint-config-next: N/A
  react: 18.3.1
  react-dom: 18.3.1
  typescript: 5.7.3
Next.js Config:
  output: N/A
 ⚠ An outdated version detected (latest is 15.2.1), upgrade is highly recommended!
   Please try the latest canary version (`npm install next@canary`) to confirm the issue still exists before creating a new issue.
   Read more - https://nextjs.org/docs/messages/opening-an-issue
```

rik-iso:
Please don't spend too many cycles on this right now if it's not trivial to reproduce (unless someone else reports the same issue of course) - I've worked around it. That said I'm happy to try and help repro it as much as possible if it helps!

I'm working around it for now by making a copy of the file that depends on `nuqs/server` and using `nuqs` in it, then prop drilling the right version of the method down depending if it's in app router or pages router 🙃 . It's highly possible there's something weird in our codebase (which is a bit of a frankenstein's monster of app and pages router at this point).

franky47:
Ahh I get the error by reproducing your exact setup (matching Node.js, React & Next.js versions). So updating some of those might solve the issue. My bet is on React, I'll see if bumping just that one to 19 solves it.

I'd like to get to the bottom of it just so that I can advise others that encounter the same issue.

rik-iso:
Just `react`? `react-dom` too?

I'm concerned that a React upgrade may be a herculean task on this old repo but iirc for the pages router it will stay effectively stuck to React 18?

but good to know. we're luckily burning the world and building a new app soon on Next 15/R19 so not inclined to go too deep on this if nuqs itself is fine (phew) and I have a workaround here...

Thanks so much for looking into this and for nuqs. Looking forward to adding nuqs to our new app too!

rik-iso:
Oh I was looking for something like pnpm catalogs just the other day, thanks for showing me that! Time to get pnpm upgraded to v10.

franky47:
You're welcome! Just watch out for postinstall scripts not running by default in v10.

franky47:
Thanks for the sponsorship! 💖

schardev:
Adding this here for anyone who checks later - thanks for the [patch](https://github.com/schardev/nuqs-issue/pull/1), it worked like a charm 👍

PR #372, opened 2023-10-22
feat: Prettify URLs

franky47:
Don't URL-encode safe characters, and fix the array parser logic.

Proposed encoding, compared with built-in solutions:
```
┌──────┬──────────────────┬────────────────┬────────────────────┐
│ char │ encodeQueryValue │ encodeURI      │ encodeURIComponent │
├──────┼──────────────────┼────────────────┼────────────────────┤
│  !   │   !              │   !            │   !                │
│  "   │          %22     │          %22   │          %22       │
│  #   │          %23     │   #            │          %23       │
│  $   │   $              │   $            │          %24       │
│  %   │          %25     │          %25   │          %25       │
│  &   │          %26     │   &            │          %26       │
│  '   │          %27     │   '            │   '                │
│  (   │   (              │   (            │   (                │
│  )   │   )              │   )            │   )                │
│  *   │   *              │   *            │   *                │
│  +   │          %2B     │   +            │          %2B       │
│  ,   │   ,              │   ,            │          %2C       │
│  -   │   -              │   -            │   -                │
│  .   │   .              │   .            │   .                │
│  /   │   /              │   /            │          %2F       │
│  :   │   :              │   :            │          %3A       │
│  ;   │   ;              │   ;            │          %3B       │
│  <   │          %3C     │          %3C   │          %3C       │
│  =   │   =              │   =            │          %3D       │
│  >   │          %3E     │          %3E   │          %3E       │
│  ?   │   ?              │   ?            │          %3F       │
│  @   │   @              │   @            │          %40       │
│  [   │   [              │          %5B   │          %5B       │
│  \   │   \              │          %5C   │          %5C       │
│  ]   │   ]              │          %5D   │          %5D       │
│  ^   │   ^              │          %5E   │          %5E       │
│  _   │   _              │   _            │   _                │
│  `   │          %60     │          %60   │          %60       │
│  {   │   {              │          %7B   │          %7B       │
│  |   │   |              │          %7C   │          %7C       │
│  }   │   }              │          %7D   │          %7D       │
│  ~   │   ~              │   ~            │   ~                │
└──────┴──────────────────┴────────────────┴────────────────────┘
```

Explanation:
- `+` is encoded as `%2B` in order not to be confused with a space according to RFC-3986. Equally, spaces will be changed to `+` (not shown here)
- `%` are encoded to avoid being interpreted as incomplete escape sequences
- `#` is encoded to avoid being interpreted as the URL hash and breaking the query string
[...]

PR #953, opened 2025-03-31
feat: Add support for TanStack Router

ahmedrowaihi:
**note** (from @franky47): the adapter in this PR is only for TanStack Router. TanStack Start support _may_ be added later when it reaches 1.0.


packages/nuqs/package.json
```diff
@@ -138,7 +143,8 @@
     "next": ">=14.2.0",
     "react": ">=18.2.0 || ^19.0.0-0",
     "react-router": "^6 || ^7",
-    "react-router-dom": "^6 || ^7"
+    "react-router-dom": "^6 || ^7",
+    "@tanstack/react-router": ">=1"
```
franky47, line 147: **suggestion**: We might want to pin it to v1 for now, as we don't know what kind of API evolutions might land in the future. 

Technically we should do this for Next too, but at least there we've got extensive testing to detect breaking changes early on.
```suggestion
    "@tanstack/react-router": "^1"
```

franky47 (changes requested):
Thank you for this! 🙌

I see the adapter is based on the next/app one, have you tried it in a TSR app? My main concern would be shallow routing support.

Also, you'll need to add the new entry points to the files list in package.json (otherwise they won't land on NPM).

franky47 (approved):
Thanks, I took the liberty of flattening the TSR adapter code, the impl file was mostly needed for Next.js to have the agnostic adapter pull out implementations from both app & pages adapters.

Looks good for a start, since TSR is mostly for SPAs, I'll add an e2e test environment on this PR tomorrow to see how it behaves. Then there's the question of how to handle the `shallow: false` option in TanStack Start for SSR, but we can deal with that later.

franky47:
The initial porting of the shared e2e test suite looks pretty good!

This is likely to be only the tip of the iceberg though, the real value in a TSR adapter would be to connect nuqs search params definition objects (the ones you feed to `useQueryStates`, `create{Loader,Serializer}` etc) to TSR's route definitions for seamless integration with type-safe routing (getting nuqs' search params in TSR's `<Link>` components, and having it use custom serialisation in the URL when doing so).

packages/nuqs/src/adapters/tanstack-router.ts
```diff
@@ -0,0 +1,45 @@
+import { useRouter, useRouterState } from '@tanstack/react-router'
+import { startTransition, useCallback, useMemo } from 'react'
+import { createAdapterProvider } from './lib/context'
+import type { AdapterInterface, UpdateUrlFunction } from './lib/defs'
+
+function useNuqsTanstackRouterAdapter(): AdapterInterface {
+  const state = useRouterState()
```
SeanCassiere, line 7: This could be optimized by using `useLocation` and subscribing only to changes to `search`.

```ts
const urlSearch = useLocation({ select: (state) => state.search })
```

franky47: Thanks for the tip, I can even transform it to a URLSearchParams object there and avoid the useMemo a few lines down:

```
const searchParams = useLocation({
  select: state => new URLSearchParams(state.search)
})
```

Asking for later, but would this also work to subscribe to a specific set of search params, for "fine grained subscriptions" as Tanner calls it? We're adding this in #855, and if it's that simple (select only the relevant keys in the slice selector), that would be beautiful.

SeanCassiere: > Edit: unfortunately it doesn't track changes if I do that, oh well.

Yea, it's best to keep that type of transformation out of the selector since it's return type is used for tracking reactivity.

> Asking for later, but would this also work to subscribe to a specific set of search params, for "fine grained subscriptions"

Yes this is how TSR handles fine-grained subscription using these selectors (basically the same concept as what y'all are working on).

snelsi:
@franky47 
While this PR is waiting to be merged, would it be possible to publish a beta version of the package to npm with this feature included for testing purposes?

franky47:
@snelsi you can install it from the pkg.pr.new link posted above.

Let me know how it works for your app!

Manjit2003:
please get this merged! waiting.....

an-h2:
going to use this through pkg.pr.new, will share any feedback I have.

AmruthPillai:
We're using this in production with the latest version of TanStack Router as of today (1.119.0) and it works perfectly, haven't found any issues so far. Thank you for all your hard work! :)

ericwang401 (approved):

ericwang401:
Sorry! I did not mean to hit approve. But great PR!

packages/e2e/tanstack-router/vite.config.js
```diff
@@ -0,0 +1,17 @@
+import { TanStackRouterVite } from '@tanstack/router-plugin/vite'
+import viteReact from '@vitejs/plugin-react'
+import { resolve } from 'node:path'
+import { defineConfig } from 'vite'
+
+// https://vitejs.dev/config/
+export default defineConfig({
+  plugins: [TanStackRouterVite({ autoCodeSplitting: true }), viteReact()],
```
SeanCassiere, line 8: Please set `target: 'react'`.

---

I know this is just CI e2e stuff, but considering in the future y'all _may_ have a bot that does auto-upgrading of dependencies, its better safe than sorry.

packages/nuqs/src/adapters/tanstack-router.ts
```diff
@@ -0,0 +1,59 @@
...
+      startTransition(() => {
+        navigate({
+          // I know the docs say to use `search` here, but it would require
+          // userland code to stitch the nuqs definitions to the route declarations
+          // in order for TSR to serialize them, which kind of breaks the
+          // "works out of the box" promise, and it also wouldn't support
+          // the custom URL encoding.
+          // TBC if it causes issues with consuming those search params
+          // in other parts of the app.
+          to: renderQueryString(search),
```
SeanCassiere, line 42: Looking at the implementation of `renderQueryString`, it looks to be outputting just a query string without an actual destination path.

In its current form, providing CI is passing on your side, TSR is kinda going YOLO and performing the update. I'm concerned that this could unintentionally break for y'all in the future, if we ever decided to touch the `router-core` logic for route-matching.

---

Could you try using this snippet to get the leaf node to be used as the `from` value in the `navigate` call?

https://github.com/TanStack/router/blob/5d940e2d8bdb12e213eede0abe8012855433ec4b/packages/react-router/src/link.tsx#L108-L112

So, it'd be added in like this.

```ts
startTransition(() => {
  // __INTERNAL_CODE_COMMENT__
  from,
  to: renderQueryString(search)
  // ...
})
```

---

If this works for y'all, I think it may be better using the `from` value instead of just a `to` value that doesn't actually point to a route in TSR.

If not, it's fine to just leave it "as-is". Its not like we touch the route-matching logic often anyways, if ever really. If it ever does break, we can always work together  to figure out a way that supports your use-case as well.

ahmedrowaihi: added from and ran the cypress myself and logged the "from" in all tests, it works good now
also shallowing is not affected

franky47: @SeanCassiere `renderQueryString` is to avoid encoding some characters in the URL, it's a bit controversial but it allows pretty-rendering things like JSON in the browser URL bar, and it gets encoded anyway when copied. See #372 for context.

Looks like CI is all green with those changes, thanks both!

SeanCassiere (approved):

dinogit:
Installed through pkg.pr.new and in production, no problem so far. Thanks for hard work.

franky47:
**note**: Linking to issue #1008: this PR will only target TanStack Router, not TanStack Start's SSR. We might update the adapter when TSS reaches 1.0, but for now it's considered out of scope.

vanquishkuso:
Has anybody tested this with basepath? I can't seem to get it to work properly. When clearing the queryState it removes the basepath as well.

For example, when writing something in a search input, it works correctly and changes the url to `http://localhost:3000/ui/customers/?search=test`, but removing everything in the search box it goes to `http://localhost:3000/customers` which results in showing the missing page component.

Basepath is set both in vite.config and createRouter to /ui. 
`const router = createRouter({
  routeTree,
  context: {
    queryClient
  },
  basepath: "/ui",
  defaultPreload: "intent",
  defaultPreloadStaleTime: 0,
  scrollRestoration: true
});`

franky47:
@vanquishkuso does the behaviour also happen if there is another search param in the URL, or only if `?search=test` is the only one?

It could be that doing `navigate({ to: '' })` is the issue here, I'll see if I can reproduce it, thanks for your feedback! 🙌

vanquishkuso:
@franky47 thank you for the fast reply!

I have five different search params in this page, one input and four select fields. When clearing them all I get the same result as above, basepath is removed and 404 page is shown. However, if i have several active and only remove one param it works as intended.

That seems to be the problem indeed, even for a regular link when I use `navigate({ to: '' })` it removes the basepath. I also see a type error `Type '""' is not assignable to type '"/" | "." | ".." |`. When using `navigate({to: '.'})` it works perfectly in my case.

vanquishkuso:
I tried to reproduce it in a sandbox but it works as intended with basepath, maybe something is wrong with my project. I will test some more.

franky47:
@vanquishkuso Thanks, could you try this preview build please? It uses `'.'` instead of an empty string when clearing the search params:

```bash
pnpm add https://pkg.pr.new/nuqs@09757c307089e26f203b391f82ba42f8cd845cbf
```

vanquishkuso:
@franky47 yes that fixes the issue!

I was also running `"@tanstack/react-router": "1.121.0-alpha.22"` and `"@tanstack/router-plugin": "^1.121.34"` originally. When i upgraded to `"@tanstack/react-router": "^1.121.34"` it worked with older `nuqs@953`. So it may have been a problem with the alpha build i was running. Super thankful for the help and thanks for the awesome work!

franky47:
Yeah I haven't been able to reproduce it locally either. The swap for `'.'` makes sense though so we can keep that, but I likely won't branch off CI to test for basepath (like we do for Next.js) unless some more issues arise with it.

franky47:
I'm doing some tests and while `validateSearch` using the `createStandardSchemaV1` utility works for most parsers, it does not for ones that don't either trivially serialise with `.toString` or in a bijective way via `JSON.stringify/parse`.

TanStack Router has a single layer of SerDe (SERialisation / DEserialisation) for **all** search params, whereas nuqs defines it per-key. By the time `validateSearch` runs, the browser's `URLSearchParams` has already gone through that layer, so `validateSearch` receives _"hydrated"_ values (strings, numbers, JSON.parse). The other issue is that TSR will rewrite search params based on the return values of `validateSearch`, going through TSR's serialisation layer.

Let's take the example of `parseAsTimestamp`:
1. Setting the state gives a URL like `?ts=1752243357912`
2. Upon reloading the page, this gets turned into a number, which is correctly interpreted as a Date object by validateSearch.
3. This goes back into the serialisation layer for search params rewrite, which uses `JSON.stringify`. The output is then `{ ts: "2025-07-11T14:15:57.912Z" }`
4. This feeds back into the `validateSearch`, which tries parsing it as an integer (timestamp), leading to only keeping `2025` as a number, turning into 2025 ms after the epoch, so a `1970-01-01T00:00:02.025Z` Date object
5. This dance repeats until the date output becomes stable at `1970-01-01T00:00:01.970Z` (note how the 1970 gets encoded both in the year and number of ms)

So yeah, kind of a mess.

The good news is this works perfectly fine if we ignore `validateSearch`, _unvalidated_ search params aren't exposed to this rewrite engine, at the expense of also not being exposed to the type-safe routing features (Link props & router calls).

Another limitation is we can't use the `urlKeys` feature for type-safe routing, as it would be ambiguous which key to use (the full variable name or the shorthand urlKey).

I'm willing to call this a limitation of compatibility between the two approaches, add a disclaimer in the docs and call it a day, but I'm also open to suggestions.

franky47:
Thanks everyone! Feel free to try the beta and leave some feedback if you see some weird things.

If everything is fine, this should ship to GA after the summer holidays. 🏝️

gensmusic:
I tried the nuqs@2.5.0-beta.3 with tanstack router and shadcn-table, I found an issue, shadcn-table sort state is not synced with search params. You have to click twice to make the search params the same with shadcn-table.
When first click on an column header, for example, 'status', you can see the search params is right ,but the column header 'status' is not changed.

image

then you click one more time, the 'status' will in the right state.

image

I test almost same code with next.js, and it works ok, so I think it's the tanstack adapter issue.

franky47:
@gensmusic thanks for the feedback, do you have a reproducible example I can try? Feel free to open a dedicated issue, thanks!

gensmusic:
> @gensmusic thanks for the feedback, do you have a reproducible example I can try? Feel free to open a dedicated issue, thanks!

Here is a demo to test with https://github.com/gensmusic/nuqs-tanstack-router-test.git
Running commands and setup process are in the README.md

ahmedrowaihi:
> I tried the nuqs@2.5.0-beta.3 with tanstack router and shadcn-table, I found an issue, shadcn-table sort state is not synced with search params. You have to click twice to make the search params the same with shadcn-table. When first click on an column header, for example, 'status', you can see the search params is right ,but the column header 'status' is not changed.
> 
> image
> then you click one more time, the 'status' will in the right state.
> 
> image
> I test almost same code with next.js, and it works ok, so I think it's the tanstack adapter issue.

feels like hydration issue

franky47:
I was able to reproduce this with just `parseAsJson`, which behaves incorrectly indeed.

Could be due to how TSR manages (de)serialisation (I see some `[object Object]` strings in the logs, more or less URL-encoded), I'll dive deeper and report back.

franky47:
Fix here: #1073.

maelp:
I'm interested by this, it's not clear to me how Tanstack Router search params methods and NUQS relate/complement each other? Are they doing the same?

By quickly reading the Tanstack Router doc I have the impression that the path suggested by Tanstack (eg when modifying filters on a table that we want to have reflected in the URL) is

input -> change URL params -> this triggers the change of state

rather than

input -> triggers a change of state -> this is synchronized (without reload) in the URL

are both Tanstack and Nuqs doing that? Am I misunderstanding something? Which would be the better way to do this and why?

Are TSR and Nuqs complementary or serving identical purposes, and one should rather choose one or the other?

franky47:
@maelp replied here: https://github.com/TanStack/router/discussions/4140#discussioncomment-14318303

gensmusic:
@franky47 Hi, I update my test repo with latest verison 2.6.0, and I still got this issue, see  https://github.com/gensmusic/nuqs-tanstack-router-test.git

If you play with the demo, you will see errors in console, I looks like it parse with a string literally "[object Object]"
image

and this is the parse code

```ts
export const getSortingStateParser = <TData>(
  columnIds?: string[] | Set<string>,
) => {
  const validKeys = columnIds
    ? columnIds instanceof Set
      ? columnIds
      : new Set(columnIds)
    : null;

  return createParser({
    parse: (value) => {
      console.log("🌹🌹 parse with value", value, `typeof value`, typeof value)
      try {
        const parsed = JSON.parse(value);
        const result = z.array(sortingItemSchema).safeParse(parsed);
        console.log("🌹🌹 parse with value result", result)

        if (!result.success) return null;

        if (validKeys && result.data.some((item) => !validKeys.has(item.id))) {
          return null;
        }

        return result.data as ExtendedColumnSort<TData>[];
      } catch (e){
        console.error("🌹🌹 parse with value error", e)
        return null;
      }
    },
    serialize: (value) => JSON.stringify(value),
    eq: (a, b) =>
      a.length === b.length &&
      a.every(
        (item, index) =>
          item.id === b[index]?.id && item.desc === b[index]?.desc,
      ),
  });
};

```
Also therer the parse is used
```ts
  const [sorting, setSorting] = useQueryState(
    SORT_KEY,
    getSortingStateParser<TData>(columnIds)
      .withOptions(queryStateOptions)
      .withDefault(initialState?.sorting ?? []),
  );
  console.log("⚽️⚽️ sorting now is", JSON.stringify(sorting))

```
And you can check the code with
https://github.com/gensmusic/nuqs-tanstack-router-test/blob/5649d57f369c09f7eea10125beb44ed616dec69d/src/hooks/use-data-table.ts#L150

I add some log and debug with local nuqs, I found this. In `packages/nuqs/src/adapters/tanstack-router.ts`  and the related code is 
```ts
  const searchParams = useMemo(
    () =>
      // search is a Record<string, string | number | object | Array<string | number>>,
      // so we need to flatten it into a list of key/value pairs,
      // replicating keys that have multiple values before passing it
      // to URLSearchParams, otherwise { foo: ['bar', 'baz'] }
      // ends up as { foo → 'bar,baz' } instead of { foo → 'bar', foo → 'baz' }
      new URLSearchParams(
        Object.entries(search).flatMap(([key, value]) => {
          if (Array.isArray(value)) {
// -- 🔥 here, sort is `[{"id":"status","desc":false}]`, it's an array of object.
            return value.map(v => [key, v])
          } else if (typeof value === 'object' && value !== null) {
            // TSR JSON.parses objects in the search params,
            // but parseAsJson expects a JSON string,
            // so we need to re-stringify it first.
            return [[key, JSON.stringify(value)]]
          } else {
            return [[key, value]]
          }
        })
      ),
    [search, watchKeys.join(',')]
  )
```

So when value is type array but its elements might be object!

franky47:
@gensmusic your `getSortingStateParser` needs to implement the same fix as in #1073. Or better, have it call nuqs' `parseAsJson` to do the (de)serialisation, and implement your filtering on top.

gensmusic:
@franky47 Oh, sorry, I just update my content above

gensmusic:
> @gensmusic your `getSortingStateParser` needs to implement the same fix as in #1073. Or better, have it call nuqs' `parseAsJson` to do the (de)serialisation, and implement your filtering on top.

Fix `getSortingStateParser` is not enough, because parser is getting a wrong string `[ojbect Object]`, but it expects a json string of `[{"id":"status","desc":false}]`

gensmusic:
> @gensmusic your `getSortingStateParser` needs to implement the same fix as in #1073. Or better, have it call nuqs' `parseAsJson` to do the (de)serialisation, and implement your filtering on top.

oh, I get it. Thanks a lot.

gensmusic:
@franky47  Hi there 👋,

Thanks again for your timely responses — and sorry for bothering you once more. I’d like to have another discussion with you about the TanStack Router adapter.

I did some research on how TanStack Router serializes and deserializes search params. According to the docs:
👉 [https://tanstack.com/router/latest/docs/framework/react/guide/custom-search-param-serialization](https://tanstack.com/router/latest/docs/framework/react/guide/custom-search-param-serialization)

By default, TSR uses something like this to handle search params:

```tsx
import {
  createRouter,
  parseSearchWith,
  stringifySearchWith,
} from '@tanstack/react-router'

const router = createRouter({
  // ...
  parseSearch: parseSearchWith(JSON.parse),
  stringifySearch: stringifySearchWith(JSON.stringify),
})
```

---

### The issue in the current Nuqs TSR adapter

In the current adapter, we only receive the **parsed search object**, but sometimes we also need to convert it back into a `URLSearchParams`.

A possible solution is to reuse TSR’s default serialization approach:

```ts
import { stringifySearchWith } from '@tanstack/react-router'

function useNuqsTanstackRouterAdapter(watchKeys: string[]): AdapterInterface {
  const search = useLocation({
    select: state =>
      Object.fromEntries(
        Object.entries(state.search).filter(([key]) => watchKeys.includes(key))
      )
  })
  ...

  // Use the same default serialization to go back to search params (Just for demonstration)
  const toSearchParams = stringifySearchWith(JSON.stringify)

  const searchParams = useMemo(
    () => new URLSearchParams(toSearchParams(search)),
    [search, watchKeys.join(',')]
  )
}
```
I tested this change locally, and it works fine. 

---

### Why I bring this up

I tested the same code across different routers (e.g., Next.js App Router), and noticed that useQueryState behaves differently depending on the router.

To reduce unexpected differences across routers, I think it would be better if the adapter tried to keep the serialization behavior as consistent as possible.

Would love to hear your thoughts on this. 🙏

ahmedrowaihi:
> @franky47 Hi there 👋,
> 
> Thanks again for your timely responses — and sorry for bothering you once more. I’d like to have another discussion with you about the TanStack Router adapter.
> 
> I did some research on how TanStack Router serializes and deserializes search params. According to the docs: 👉 https://tanstack.com/router/latest/docs/framework/react/guide/custom-search-param-serialization
> 
> By default, TSR uses something like this to handle search params:
> 
> ```tsx
> import {
>   createRouter,
>   parseSearchWith,
>   stringifySearchWith,
> } from '@tanstack/react-router'
> 
> const router = createRouter({
>   // ...
>   parseSearch: parseSearchWith(JSON.parse),
>   stringifySearch: stringifySearchWith(JSON.stringify),
> })
> ```
> 
> ### The issue in the current Nuqs TSR adapter
> In the current adapter, we only receive the **parsed search object**, but sometimes we also need to convert it back into a `URLSearchParams`.
> 
> A possible solution is to reuse TSR’s default serialization approach:
> 
> ```ts
> import { stringifySearchWith } from '@tanstack/react-router'
> 
> function useNuqsTanstackRouterAdapter(watchKeys: string[]): AdapterInterface {
>   const search = useLocation({
>     select: state =>
>       Object.fromEntries(
>         Object.entries(state.search).filter(([key]) => watchKeys.includes(key))
>       )
>   })
>   ...
> 
>   // Use the same default serialization to go back to search params (Just for demonstration)
>   const toSearchParams = stringifySearchWith(JSON.stringify)
> 
>   const searchParams = useMemo(
>     () => new URLSearchParams(toSearchParams(search)),
>     [search, watchKeys.join(',')]
>   )
> }
> ```
> 
> I tested this change locally, and it works fine.
> 
> ### Why I bring this up
> I tested the same code across different routers (e.g., Next.js App Router), and noticed that useQueryState behaves differently depending on the router.
> 
> To reduce unexpected differences across routers, I think it would be better if the adapter tried to keep the serialization behavior as consistent as possible.
> 
> Would love to hear your thoughts on this. 🙏

please do a PR if you can

franky47:
Yeah let's move this to a new PR, there are a couple of things we could test against too, like custom `parseSearchWith` / `stringifySearchWith` pairs (eg: base64 encoding everything). Thanks @gensmusic !

PR #961, opened 2025-04-03
test: Add property-based testing with fast-check

franky47:
Thanks @dubzzz! 🙌

### DIFF
--- a/packages/nuqs/vitest.setup.ts
+++ b/packages/nuqs/vitest.setup.ts
@@ -1,8 +1,15 @@
 import * as matchers from '@testing-library/jest-dom/matchers'
 import { cleanup } from '@testing-library/react'
+import fc, { VerbosityLevel } from 'fast-check'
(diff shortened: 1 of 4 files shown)


packages/nuqs/vitest.setup.ts
```diff
@@ -1,8 +1,15 @@
...
 
 expect.extend(matchers)
 
 // https://testing-library.com/docs/react-testing-library/api/#cleanup
 afterEach(cleanup)
+
+fc.configureGlobal({
+  numRuns: 1000,
+  verbose: process.env.CI ? VerbosityLevel.None : VerbosityLevel.VeryVerbose,
+  interruptAfterTimeLimit: 5000
```
dubzzz, line 14: Probably worth putting slightly below the time limit of Vitest otherwise fast-check will never be able to cut on time for properties taking very long to assess.

PR #963, opened 2025-04-04
chore(docs): Waku v0.22 community adapter useRouter update 

rmarscher:
useRouter is no longer unstable in Waku since v0.22.

### DIFF
--- a/packages/docs/content/docs/community-adapters/waku.mdx
+++ b/packages/docs/content/docs/community-adapters/waku.mdx
@@ -20,7 +20,7 @@ import {
   unstable_createAdapterProvider as createAdapterProvider,
   renderQueryString,
 } from "nuqs/adapters/custom";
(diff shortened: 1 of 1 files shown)


franky47 (approved):
Thanks! Maybe we could add a comment about how to import it in older versions?

Since community adapters don't have SemVer ranges enforced in a package.json, I'm not sure what's the best strategy to deliver them over time:
- Always target latest
- Have a semver range added as a comment at the top of the file
- Other ideas?

rmarscher:
Thanks! I added a comment and edited the title of this PR to include that waku version number. I think most people currently running Waku are following pretty closely as the API has been unstable pre-1.0. Targeting the latest and letting people look at the community adapter git history is probably fine. I don't expect it to change moving forward so this shouldn't be too much of an issue.

PR #977, opened 2025-04-14
chore: Update dependencies

franky47:


fitimbytyqi:
@franky47 I think this PR introduced an error in the docs page https://nuqs.47ng.com/docs/installation, since it was working before ?

franky47:
@fitimbytyqi I thought it was the Sentry SDK update to v9, but it turns out it was the `next build --turbopack` from #971 that Sentry does not yet support. Sorry about the outage, it's resolved now.

That'll teach me to have different staging & prod environments 😅

Issue #691, opened 2024-10-22
Error while loading rule 'import/no-cycle': Package subpath './adapters/next/app' is not defined by "exports"


GBvanDam:
## Context

What's your version of `nuqs`? 

```
"nuqs": "^2.0.0",
```

What framework are you using?

- ✅ Next.js (app router)

Which version of your framework are you using?

```
"next": "^14.2.15",
```

## Description

eslint fails on `import { NuqsAdapter } from 'nuqs/adapters/next/app'` with message: 

```bash
Error while loading rule 'import/no-cycle': Package subpath './adapters/next/app' is not defined by "exports"
```

## Reproduction

1. `import { NuqsAdapter } from 'nuqs/adapters/next/app'`
2. install `"eslint": "^8.57.0"`
3. install `"eslint-plugin-import": "^2.31.0"`
4. install `"eslint-import-resolver-alias": "^1.1.2"`
5. run eslint

franky47:
Thanks for the report, the bundling looks fine on my end. What's your ESLint version?

Could you paste your tsconfig.json here please?

franky47:
Ah I think I see what the problem might be.

For some TypeScript configs (that didn't support/honor export paths), the import needed to have top-level .d.ts files that point to the correct ones in the `dist` directory.

Looks like the ESLint rule chokes on those files, where: `node_modules/nuqs/adapters/next/app.d.ts` exports everything from `node_modules/nuqs/dist/adapters/next/app.d.ts`.

I'll see in what cases we need those intermediate top-level declaration files. If we could do without, I'd be happy, they are a pain to maintain.

acocheo:
Almost same problem here I think. Same eslint `^8.57.0` but happens also  with `v.^2.0.1` 
```
 Error while loading rule 'import/no-cycle': Package subpath './adapters/next/app' is not defined by "exports" in /[...]/node_modules/nuqs/package.json
 Occurred while linting [...]
    at exportsNotFound (node:internal/modules/esm/resolve:304:10)
    at packageExportsResolve (node:internal/modules/esm/resolve:594:13)
    at resolveExports (node:internal/modules/cjs/loader:634:36)
    at Module._findPath (node:internal/modules/cjs/loader:724:31)
    at findModulePath ([...]/node_modules/eslint-import-resolver-alias/index.js:99:27)
    at exports.resolve ([...]/node_modules/eslint-import-resolver-alias/index.js:75:10)
[...]

PR #996, opened 2025-05-12
fix: Replace require by default conditional export field

stefan-schubert-sbb:
This is useful for transforming to reach esm module by default.


franky47:
**question** Wouldn't the import conditions algorithm have already resolved to either `import` (from ESM) or `require` (from CJS) by the time it reaches this `default` condition?

What are cases where this would not be the case?

stefan-schubert-sbb:
I retested this code change again -> now it is broken... the package loader doesn't reach the default field, because require field exists. At the test before I created this PR I replaced the require field by the default field using the esm *.js file -> then all is fine. Could this be a solution providing a default field with the esm export instead a cjs export which throws an error.

stefan-schubert-sbb:
I updated the change...

stefan-schubert-sbb:
> I tested this PR against the repro for #691 (in https://github.com/franky47/nuqs-repro-691), and it passes there so that's good.
> 
> > The thrown error is now not really nice to read
> 
> @stefan-schubert-sbb What do you mean by that?

Your cjs module produces a good error message: "This package is ESM only...". If we provide the nuqs esm package through default export field, it will produce the error "SyntaxError: Cannot use import statement outside a module", if it is not transpiled first. This should be interpreted right by the developer. In addition a faq related to this error message could be provided.

franky47:
>  If we provide the nuqs esm package through default export field, it will produce the error "SyntaxError: Cannot use import statement outside a module", if it is not transpiled first.

That makes sense, since the CJS `require` resolution would end up being fed an ESM module. The `esm-only.cjs` thing is a hack (which some package linters are rightfully complaining about), but it sends a clear message. This is why I'm not super keen on widening the `default` resolution for CJS, since we dropped support for it in v2.0.0.

Now, since Node.js backports `require(esm)` in all actively supported versions (>=20), it feels like keeping this package pure ESM should work without additional transpilation, what was the use-case where you needed this change?

stefan-schubert-sbb:
The usecase is testing under nextjs with jest. I think to provide the esm package as default is the right way, instead being broken for transpilers using a cjs module and throwing an error. Nowadays developers must be careful using up to date npm packages in a cjs environment without transpiling. I think it would be a good compromise to use the default field instead of the require field to support both usecases without being broken for one.

TkDodo:
I tried out this PR preview build in the sentry code-base to make our jest tests work (because it currently always picks up the `esm-only.cjs`), and it works great.

With this, I can just run `nuqs` through babel and the tests pass without errors.

Right now, I need a `moduleNameMapper` to force using `dist/index.js`:

```
moduleNameMapper: {
  '^nuqs$': '<rootDir>/node_modules/nuqs/dist/index.js',
}
```

which isn’t that nice.

franky47:
This makes a lot of sense now that:
- `require(esm)` has been backported/unflagged in Node.js 20+
- Node.js 18 is deprecated

The existing "crash with a message saying that this is ESM only" prevents `require(esm)` doing its job, whereas this change lets Node.js run fine.

I used the following test file to force Node.js in CJS mode, which crashed with nuqs@latest but passed fine with this preview.
```js
// main.cjs
const { createLoader, parseAsInteger } = require('nuqs')

const searchParams = {
  foo: parseAsInteger.withDefault(0),
}

const load = createLoader(searchParams)

console.dir(load('?foo=42&bar=123'))
```

I'll add a few changes to edit/remove the "ESM only" error docs and this should be good to merge.

PR #997, opened 2025-05-14
feat: add astro adapter

LenghakHok:
- Add astro adaptor that has context state management of [NanoStores](https://github.com/nanostores/nanostores)


franky47:
Thanks, but one issue I see with this is adding two dependencies for everyone, that only makes sense for Astro. Is there a way we could use nanostores scoped only to an Astro-specific adapter (and therefore not have those two dependencies bundled with the core code in other frameworks)?

LenghakHok:
Thank you for your feedback. I understand the concern about introducing two additional dependencies for everyone. However, given their minimal sizes — `nanostores (265–803 bytes minified and brotlied)` and `@nanostores/react (less than 1 KB)`, totaling around 1–1.2 KB — I am afraid that I currently don’t see an alternative that would allow us to use `nanostores` exclusively for Astro-scoped projects.

Most changes are contained within the core adapter context file. Making the solution framework-scoped could require additional conditional checks and may introduce breaking changes for developers.

Alternatively, separating each adapter into its own package could offer more flexibility regarding customization and dependencies.

This is the limitation I’m facing. I suggest we consult with more experienced developers to explore potential solutions.

LenghakHok:
Let me try "astro only env" as you have mentioned, and already has some codes done! I will need to create an e2e testing for astro as well! Might need more of your review on the way, too!

LenghakHok:
@franky47 My solution for now is to:
- Create a custom context that uses nanostores
- Create an astro adapter based on a custom react adapter that use only nanostores and astro-only env
- Extract global logic of useQueryState and useQueryStates to core folder while maintaining types exports and core functionality without no breaking changes to other adapters
- Create another useQueryState and useQueryStates for astro only that uses the nanostore context

Do you think this is a good solution?

franky47:
I think I prefer having the nanostore dependency for everyone over maintaining multiple implementations of the core hook logic, even if separated like this.

The main reason behind the adapter architecture in v2 was to allow building framework-agnostic components that use the same imports, and can be reused in any of the compatible frameworks.

Coming back to nanostores, could this also be used to implement fine grained subscriptions / key isolation (see #855)? I managed to make it work for all adapters but Next.js, which uses a single Context for propagating the URLSearchParams object, and therefore causing all useSearchParams to re-render when only one of them actually changes.

LenghakHok:
Very interesting approach! I believe managing state with NanoStores is a great idea, as nanostores is framework-agnostic by default, making it easier to scale. 

Regarding fine-grained reactivity, nanostores fully supports it through features like [Deep Map](https://github.com/nanostores/nanostores?tab=readme-ov-file#deep-maps).

My initial proposal was to use NanoStores solely for the context object. However, we could take this further by integrating Deep Map into useQueryState and useQueryStates, while continuing to support a framework-agnostic architecture.

I'll take a deeper look into the implementation of our core hooks to explore how we might integrate them. In the meantime, if you have a moment, it'd be great if you could take a quick look at Nanostores — you already have a much clearer vision for how they could fit into the architecture.

LenghakHok:
I will close one for now and add a new pr instead.

LenghakHok:
Nanostores now appear to be the most viable solution. The migration would proceed as follows:

Replace independent internal state with a global object state using Nanostores’ deepMap.

On full URL changes (i.e., page reloads), update the deepMap with all query parameters as a unified global object.

Since deepMap provides fine-grained reactivity by default, we only need to migrate useQueryState and useQueryStates to interact with and update the global object.

PR #1001, opened 2025-05-26
doc: Update server-side.mdx

ruchernchong:
Update docs for Next.js API routes to that uses the `searchParams` correctly from `NextRequest`

### DIFF
--- a/packages/docs/content/docs/server-side.mdx
+++ b/packages/docs/content/docs/server-side.mdx
@@ -85,9 +85,10 @@ loadSearchParams(new URLSearchParams(...))
 ```
 
 ```tsx tab="API routes"
+import { NextRequest } from "next/server";
 // App router, eg: app/api/location/route.ts
-export async function GET(request: Request) {
-  const { latitude, longitude } = loadSearchParams(request)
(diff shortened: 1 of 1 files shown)


ruchernchong:
> The code example was for the app router route handlers, which do use the built-in Request object type:
> 
> https://nextjs.org/docs/app/building-your-application/routing/route-handlers#convention
> 
> Your example seems to be for the pages router, we could add that too but it would be better to make the distinction in the tab title.

I am using the App Router one. I just tested on one of my projects and to get it to work, I had to make the change as per what I changed in this PR.

franky47:
That's strange, we do have a test for this which runs fine with the API showcased in the docs:

https://github.com/47ng/nuqs/blob/8bfa521041e50874939580ef86004f56f6107a61/packages/e2e/next/src/app/api/app/loader/route.ts

Would you be able to open an issue with your detailed setup and a reproduction please?

franky47:
Just with that description I cannot see a difference from what we do in the test (apart from the API route output, which I assume is irrelevant).

ruchernchong:
Closing this for now.

PR #1007, opened 2025-06-05
feat(context): moved to nanostore

LenghakHok:
- Migrate from react context to nanostores for context for now


Lenghak:
This PR is not including key-isolation feature, yet.

Lenghak:
This pr will add support for astro! However, an alternative has already added to nanostores! I just noticed this in the official repo https://github.com/VdustR/nanostores-qs.

PR #1025, opened 2025-06-24
docs: Extend next-app example to include more features

I-3B:
This PR add on the basic next-app example to use more features that would help Next.js's users to utilize the library better:
- search params cache
- serializer
- useQueryStates


I-3B:
@franky47 would you please give this a review?
I wanted to cover more features that are relevant to Next.js, while trying not to make it too complex too

PR #1028, opened 2025-06-27
Async parsers

brhx:


franky47:
Parsers can't be async, as they need to be side-effect free according to the docs. This is because the parse and serialize methods need to be mirrors of each other, and can be called at arbitrary moments in time.

If you need to perform data fetching, you should do so in a component or a custom hook that depends on the result of a sync query state.

brhx:
@franky47 

async !== side effects
sync !== no side effects

You may want to use `await crypto.subtle`, for example to encrypt your search params which isn't possible with sync-only

brhx:
That said, I can't figure out a nice way to make async parsers work without type-level breaking changes so I'll come back to it

PR #1036, opened 2025-07-04
feat: parseAsTuple

### LINKED ISSUE #1022: parseAsTuple
### Discussed in https://github.com/47ng/nuqs/discussions/1019

Please see the discussion there for guidance if you want to pick this issue for a PR. 🙏

<sup>Originally posted by **tacomanator** June 20, 2025</sup>
There have been a few situations where this would be useful, so I wanted to suggest it.

For example, a date range parser could be implemented as:

```ts
parseAsTuple([parseAsIsoDate, parseAsIsoDate])
```

Advantages over `parseAsArrayOf(parseAsDate)`:

1. Ensures both dates are present.
2. Resulting type of `[string, string]` instead of `string[]` prevents certain type errors.
3. Elements could have different types (not relevant in this example).

Just an example there are obviously many others.</div>

I-3B:
closes #1022

### DIFF
--- a/packages/nuqs/src/parsers.test.ts
+++ b/packages/nuqs/src/parsers.test.ts
@@ -16,7 +16,8 @@ import {
   parseAsString,
   parseAsStringEnum,
   parseAsStringLiteral,
-  parseAsTimestamp
+  parseAsTimestamp,
+  parseAsTuple
 } from './parsers'
 import {
   isParserBijective,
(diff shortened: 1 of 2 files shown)


I-3B:
@franky47 can you please check the interface design? if it looks good I'll proceed with tests.

franky47:
The overall design looks good, but we're blowing up the size limit budget. I need to do some shrinking of the bundle size before we can add any more features in the package (ideally I'd like to keep it under 5kB but that might be tricky to maintain).

I-3B:
> The overall design looks good, but we're blowing up the size limit budget. I need to do some shrinking of the bundle size before we can add any more features in the package (ideally I'd like to keep it under 5kB but that might be tricky to maintain).

I see
Anything I can do from my side for now?

franky47:
Hey, coming back to this PR, I'm considering adding support in the nuqs docs for the [shadcn CLI](https://ui.shadcn.com/docs/registry/getting-started), to install custom parsers with a one-line command.

Would you mind if we moved this parser to the community section, and have it installable this way?

I'd like to keep the core bundle size small, and this could be one way.

I-3B:
> Hey, coming back to this PR, I'm considering adding support in the nuqs docs for the [shadcn CLI](https://ui.shadcn.com/docs/registry/getting-started), to install custom parsers with a one-line command.
> 
> Would you mind if we moved this parser to the community section, and have it installable this way?
> 
> I'd like to keep the core bundle size small, and this could be one way.

of course, would you like me to assist with that?

franky47:
That would be great, thanks!

I haven't looked too deep into the CLI docs yet, but I saw Shadcn released a v3 that includes namespaces. From what I understand it could give us nice install commands like:

```shell
$ npx shadcn add @nuqs/parsers/parseAsTuple
# or even shorter (at the risk of feature mixup):
$ npx shadcn add @nuqs/parseAsTuple
```
However it would require declaring that namespace and point it to our registry in the components.json config, so that's an extra step.

We could go for plain URLs too:
```shell
$ npx shadcn add https://nuqs.47ng.com/parsers/parseAsTuple
```

What do you think?

I-3B:
the first command looks nicer for sure, but i don't think there's a real UX difference, as users will most likely copy-paste it from docs either ways

so i think it just depends on how much of a pain it's to maintain the registry (which i'm not knowledgeable enough about)

franky47:
> the first command looks nicer for sure, but i don't think there's a real UX difference, as users will most likely copy-paste it from docs either ways

Yeah I'd expect a single copy-paste to be enough to get started, but if you need to declare the registry first, that's an extra step that might defeat the purpose of using this CLI. TBC.

I-3B:
> Yeah I'd expect a single copy-paste to be enough to get started, but if you need to declare the registry first, that's an extra step that might defeat the purpose of using this CLI. TBC.

okay, but how would it work, let's say user executes this:
```bash
npx shadcn add https://nuqs.47ng.com/parsers/parseAsTuple
```
would this add a components/nuqs/parseAsTuple.ts file that people can import into their project directly?
or should follow a different structure

franky47:
We can now have trusted namespaced registries, see #1118.

This is one of the reasons I migrated to nuqs.dev this weekend, so we can have a nice URL as a base.

I-3B:
I think this PR will be on hold till the registry is setup and merged? as this will need:
- to be moved out of `nuqs` package to where the registry package (?).
- add it to `registry.json`

PR #1038, opened 2025-07-06
feat(types): Add `const` modifier to literal parsers to auto-infer their arguments as literals

neefrehman:
I often have to add the `as const` modifier to arrays of strings or numbers to get the types to work properly, as is also suggested by the docs. I thought it was worth opening this PR that adds [a const modifier](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-0.html#const-type-parameters) to the type parameters for the literals, so that it can automatically do this. I've also updated a handful of examples and docs to illustrate the simpler usage.

Hopefully the you agree that it's a nice type enhancement @franky47, but feel free to close this PR if not :)


packages/docs/content/docs/parsers/built-in.mdx
```diff
@@ -139,9 +139,19 @@ These parsers extend the basic integer and float parsers, but test against
...
 
+```ts
+import { parseAsStringLiteral, inferParserType } from 'nuqs'
+
+// Create parser
+const parser = parseAsStringLiteral(['asc', 'desc'])
+
+// Optional: extract the type
+type SortOrder = inferParserType<typeof parser> // 'asc' | 'desc'
+```
```
neefrehman: I wasn't sure which type of usage you'd want to take precedence in the docs here, so I'm just adding the new-style usage before the `as const` one (both here and for `parseAsNumberLiteral`). Let me know what you want to put here and I can update it.

franky47: I think both cases are valid and depend on user preference:
- Using the parser as the source of truth (and deriving types with `inferParserType`). Caveat being that you can't iterate over the values at runtime.
- Using a separate declaration for the values (which requires `as const`, important to keep in the docs), and do the `typeof values[number]` trick for union type inference

franky47:
I had read about that const modifier in generics when it came out but it didn't occur to me to use it here, thanks for the suggestion, it's a nice improvement!

franky47 (approved):

packages/docs/content/docs/parsers/built-in.mdx
neefrehman: Ok great, sounds like having both is good then 🙏🏾

PR #1041, opened 2025-07-08
doc: Fix transition docs to not call parser as function

phelma:
Docs appear to be incorrect, parsers aren't functions that can be called like this


franky47 (approved):
Oh, good catch, thanks!

Issue #798, opened 2024-12-09
Monorepo setup - Error: [nuqs] nuqs requires an adapter to work with your framework.


janglad:
## Context

What's your version of `nuqs`?
2.2.3

```
"nuqs": "^2.2.3",
```
What framework are you using?

Next JS 

- ✅/❌ Next.js (app router) (with React Compiler turned on, using Turbopack in dev but issue also occurs during build)

Which version of your framework are you using?

```
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.0.0: Mon Aug 12 20:51:54 PDT 2024; root:xnu-11215.1.10~2/RELEASE_ARM64_T6000
  Available memory (MB): 65536
  Available CPU cores: 10
Binaries:
  Node: 22.10.0
  npm: 10.9.0
  Yarn: 1.22.22
  pnpm: 9.12.2
Relevant Packages:
  next: 15.0.3 // There is a newer version (15.0.4) available, upgrade recommended!
  eslint-config-next: N/A
  react: 19.0.0-rc-fb9a90fa48-20240614
  react-dom: 19.0.0-rc-fb9a90fa48-20240614
  typescript: 5.6.3
Next.js Config:
```

## Description

We have a turbo pnpm monorepo with next js. Nuqs has been acting up for a bit, erroring out and saying that the adapter isn't installed. This is however the case (and the adapter gets resolved fine). In the past this was fixed by uninstalling nuqs from the pacakges that used it and adding it again, however this now also isn't working. Below is relevant information about nuqs in the pnpm-lock

```
importers:
  apps/web:
    dependencies:
      ...
      nuqs:
        specifier: ^2.2.3
        version: 2.2.3(next@15.0.3(@babel/core@7.25.2)(@opentelemetry/api@1.9.0)(@playwright/test@1.48.2)(babel-plugin-react-compiler@19.0.0-beta-8a03594-20241020)(react-dom@19.0.0-rc-fb9a90fa48-20240614(react@19.0.0-rc-fb9a90fa48-20240614))(react@19.0.0-rc-fb9a90fa48-20240614))(react@19.0.0-rc-fb9a90fa48-20240614)
```

```
importers:
  packages/some-package-that-is-used-in-web:
    dependencies:
      ...
      nuqs:
        specifier: ^2.2.3
[...]

PR #1046, opened 2025-07-17
Update NUQS-404.md

dmytro-palaniichuk:
Fixed reference to monorepo issue with nuqs
https://github.com/47ng/nuqs/issues/798


franky47:
Oh good catch thanks!

Merging this from the Eurostar under the English Channel 🚅🇫🇷🇬🇧😎

PR #1057, opened 2025-08-12
fix: Type inference for defaultValue of object syntax

TkDodo:
The problem was that when passing an arbitrary `string` to `defaultValue` when a `parser` was defined, the wrong overload was picked up because we basically “fall through” to the overload where we have `defaultValue: string`.

Now, that last, basic overload is intersected with:

```
parse?: never
serialize?: never
eq?: never
```

so that it’s not picked up when a parser is specified

### DIFF
--- a/packages/nuqs/src/parsers.ts
+++ b/packages/nuqs/src/parsers.ts
@@ -4,10 +4,6 @@ import { safeParse } from './lib/safe-parse'
 
 type Require<T, Keys extends keyof T> = Pick<Required<T>, Keys> & Omit<T, Keys>
 
-export type NoParser = {
-  [K in keyof Parser<unknown>]?: never
-}
-
 export type Parser<T> = {
   /**
    * Convert a query string value into a state value.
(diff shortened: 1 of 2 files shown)


packages/nuqs/src/parsers.ts
```diff
@@ -4,6 +4,10 @@ import { safeParse } from './lib/safe-parse'
 
 type Require<T, Keys extends keyof T> = Pick<Required<T>, Keys> & Omit<T, Keys>
 
+export type NoParser = {
+  [K in keyof Parser<unknown>]?: never
+}
```
franky47: **nitpick**: Everything exported by parsers.ts is also barrel-exported by `'nuqs'` and `'nuqs/server'`, so this will be available in userland. It might be necessary for TypeScript to resolve types, but adding a little JSDoc to explain why we need this might help me remember why it's there in 6 months 😅

TkDodo: In that case I think we should inline the type so that we don't expose it to consumers

franky47:
Thank you! 🫶

PR #1066, opened 2025-08-14
feat: defaultOptions for NuqsAdapter

TkDodo:
This PR adds `defaultOptions` to `<NuqsAdapter>`, which allows to override the globally specified defaults.

As a first step, only `shallow?: boolean` can be provided, but it can be extended to more default values.

### DIFF
--- a/packages/docs/content/docs/options.mdx
+++ b/packages/docs/content/docs/options.mdx
@@ -32,6 +32,16 @@ setState('foo', { scroll: true })
 
 Call-level options will override hook level options.
 
+Default values for some options can also be configured globally via the `defaultOptions{:ts}`
+[adapter](/docs/adapters) prop:
+
+```tsx
+// [!code word:defaultOptions]
+<NuqsAdapter defaultOptions={{ shallow: false }}>
+  {children}
+</NuqsAdapter>
+```
+
 
 ## History
 


packages/nuqs/src/adapters/lib/context.ts
```diff
@@ -52,10 +58,10 @@ export type AdapterProvider = ({
 export function createAdapterProvider(
   useAdapter: UseAdapterHook
 ): AdapterProvider {
-  return ({ children, ...props }: { children: ReactNode }) =>
+  return ({ children, defaultOptions, ...props }) =>
     createElement(
       context.Provider,
-      { ...props, value: { useAdapter } },
+      { ...props, value: { useAdapter, defaultOptions } },
```
TkDodo, line 64: I am not sure spreading `...props` in the position it is now currently in is correct. Looking at the types, after taking out `children` and `defaultOptions`, it’s just an empty objet `{}` so it can’t really be anything ?

Also, a context provider doesn’t have any other prop than `value`, so I was thinking the better way would probably be this:


```suggestion
  return ({ children, ...props }) =>
createElement(
context.Provider,
      { value: { useAdapter, ...props } },
```

that way, we keep `defaultOptions` inside `props` and just spread it next to `useAdapter` into `value`

franky47: Yes I came into that issue too, if we use that context to pass adapter props down to the hooks it needs to go in the value.

packages/nuqs/src/adapters/lib/context.ts
```diff
@@ -11,7 +11,13 @@ import { debugEnabled } from '../../lib/debug'
 import { error } from '../../lib/errors'
 import type { AdapterInterface, UseAdapterHook } from './defs'
 
-export type AdapterContext = {
+export type AdapterDefaultOptions = {
```
franky47: **nitpick**: There are other things we could pass through that Context from adapter props, so this type could be more generically named `AdapterProps`, what do you think?

franky47 (changes requested):
Thanks, it looks good, I'd be more generic about the `AdapterDefaultOptions` type to allow extending it in the future with other props passed via the adapter (like a callback function to process URLSearchParams after applying updates for alphabetical sorting).

I don't think it's an issue, but could passing an object from the render body cause re-renders? I guess as long as the parent component of `<NuqsAdapter>` isn't re-rendered it shouldn't be an issue, and that would cause the whole app to re-render anyway (as adapters are placed high in the tree).

```tsx
function RootLayout({ children }) {
  return (
    <NuqsAdapter defaultOptions={{ shallow: false })>
      {children}
    </NuqsAdapter>
  )
}
```

TkDodo:
@franky47 regarding re-renders: currently, the context `value` is not memoized - it’s a new object with `useAdapter` inside:

https://github.com/47ng/nuqs/blob/aa902c2ca2745f3d491b527ae5b728ec047183e1/packages/nuqs/src/adapters/lib/context.ts#L42

so that alone could cause re-renders. However, that’s only a concern IF the `RootLayout` in your example would render “from within itself”, because then `children` would be stable.

If there’s a top-down render, `children` will be new and will therefore always re-render the whole tree downwards.

packages/nuqs/src/adapters/lib/context.ts
TkDodo: yeah that makes a lot of sense. `AdapterDefaultOptions` is actually the `defaultOptions` object inside the `AdapterProps` 👍 

5025028d

Multiply:
> (like a callback function to process URLSearchParams after applying updates for alphabetical sorting).

Is this feature planned? We're definitely looking for this feature specifically. (alphabetical sorting of parameters)
Edit: or a callback where we can do it ourselves.

franky47:
@Multiply would you like to try your hand at a PR?

Multiply:
> @Multiply would you like to try your hand at a PR?

Was already looking into the recent introduction of global default config.
Do you have a preference for where the middleware should be called?

franky47:
Let's continue the discussion in #1109.

PR #1068, opened 2025-08-15
fix: Add encoding for additional special characters in encodeQueryValue

kimjeongwonn:
# Changes

Add {}|\^? to the list of special characters that are encoded in the encodeQueryValue function.

# Motivation

According to RFC 3986, certain special characters that are not included in the allowed URI character set can cause some nginx servers to return errors, making the site inaccessible when these characters appear in query parameters.
To address this issue, we are adding encoding for {}|\^? characters. While I understand that PR #372 implemented custom encoding rules instead of using encodeURIComponent to keep URLs more readable, I believe proper functionality should take priority over aesthetic considerations.

Additionally, the reason for encoding ? is that some websites use ? as a delimiter to split strings in their logic. If ? characters in query values are not encoded, it could lead to unexpected parsing issues. For safety and reliability, we should encode ? characters as well.


franky47:
Do you have an example nginx setup (version + config) where this occurs that I can try?

Anyway, this would be a breaking (behavioural) change so it will only land in nuqs@3.0.0.

felixscheinost:
We have the same issue with `[]` and Tomcat. Tomcat doesn't allow `[` and `]` in the query part unless `relaxedQueryChars` is configured to allow it.

I think `[` and `]` should also be encoded to `%5B` abd `%5D` respectively.

PR #1070, opened 2025-08-15
doc: prevent NaN appearing in pagination example

87xie:
a small fix to prevent `NaN` from appearing in the URL query. When the toggled state is off, the `onValueChange` callback receives an empty string, which is parsed as `NaN`.


franky47 (approved):
Thanks!

PR #1074, opened 2025-08-18
chore: Configure MDX types

remcohaszing:
- Augment the `JSX` namespace in `mdx/types.js` to make it understand React.
- Make all `.mdx` files part of the TypeScript program by including them in `tsconfig.json`.
- Remove arguments from `useMDXComponents()`. This project called the function manually with a stub argument, but normally this function doesn’t receive any arguments.
- Move components outside of `useMDXComponents()`, so that no React components are defined during render.
- Define the `MDXProvidedComponents` type globally, so you get TypeScript support for injected components inside MDX files.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -9,11 +9,11 @@
     },
     {
       "label": "Docs",
-      "query": "repo:47ng/nuqs is:pr is:open label:documentation"
+      "query": "repo:47ng/nuqs is:pr is:open label:documentation draft:false"
     },
     {
-      "label": "All",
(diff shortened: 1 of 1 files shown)


remcohaszing:
You can find more information here: https://github.com/mdx-js/mdx-analyzer#use

For example, you can turn those subtle type hints into red squiggly errors by enabling `mdx.checkMdx`.

remcohaszing:
Also sadly `tsc` doesn’t support plugins. This is only for editor support. I’m low-key looking into this.

fuma-nama:
It's new to me, `MDXProvidedComponents` would be a useful thing

PR #1079, opened 2025-08-19
feat: Add support for more global default options at the adapter level

franky47:
Supported:
- `shallow` defaults to `true`, change to `false` to always send 
notifications to the server on updates
- `scroll` defaults to `false`, change to `true` to always scroll to
the top of the page after an update
- `clearOnDefault` defaults to `true`, change to `false` to always
write values to the URL, even when set to the default value (explicit defaults)
- `limitUrlUpdates` defaults to a custom throttle rate for your browser,
increase it here if need be (debouncing at the top level is not recommended).

cc @TkDodo

### DIFF
--- a/packages/docs/content/docs/options.mdx
+++ b/packages/docs/content/docs/options.mdx
@@ -37,7 +37,14 @@ Default values for some options can also be configured globally via the `default
 
 ```tsx
 // [!code word:defaultOptions]
-<NuqsAdapter defaultOptions={{ shallow: false }}>
+<NuqsAdapter
+  defaultOptions={{
+    shallow: false,
+    scroll: true,
+    clearOnDefault: false,
+    limitUrlUpdates: throttle(250),
+  }}
+>
   {children}
 </NuqsAdapter>
 ```
--- a/packages/nuqs/src/adapters/lib/context.ts
+++ b/packages/nuqs/src/adapters/lib/context.ts
@@ -7,14 +7,15 @@ import {
   type ReactElement,
   type ReactNode
 } from 'react'
+import type { Options } from '../../defs'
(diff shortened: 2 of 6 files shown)


packages/nuqs/src/adapters/react.ts
```diff
@@ -93,15 +93,18 @@ const NuqsReactAdapter = createAdapterProvider(useNuqsReactAdapter)
...
   children: ReactNode
   fullPageNavigationOnShallowFalseUpdates?: boolean
 }): ReactElement {
   return createElement(
     NuqsReactAdapterContext.Provider,
     { value: { fullPageNavigationOnShallowFalseUpdates } },
-    createElement(NuqsReactAdapter, null, children)
+    // @ts-expect-error AdapterProvider expects children in its props type,
+    // but we pass them as the third argument to createElement, not via props.
+    createElement(NuqsReactAdapter, adapterProps, children)
```
TkDodo, line 107: would it work to just pass `children` as part of `adapterProps` and omit the 3rd param to `createElement` ? The `ts-expect-error` will “swallow” all other type errors that might come up in the future, too.

alternatively, `adapterPrps` could contain `children:null` that would get overwritten by the 3rd argument to `createElement` (I guess)

franky47: From looking at [how `createElement` works](https://github.com/facebook/react/blob/3e20dc8b9c4d335c77e246d114a1bcfd1b77aaae/packages/react/src/jsx/ReactJSXElement.js#L613-L647), either way seems fine, the config assigns its `children` property to the element props, then anything from the `children` onwards (can be variadic) gets merged onto that. It's a shame there are two ways to do the same thing, but oh well.

Fix in #1080.

PR #1080, opened 2025-08-20
chore: Pass children as config in createElement to avoid ts-expect-error

franky47:
From feedback by @TkDodo in https://github.com/47ng/nuqs/pull/1079#discussion_r2287880913.


TkDodo (approved):
this is great 👍

Issue #1075, opened 2025-08-18
Type-check & build docs in CI


franky47:
Add a CI step that:
1. Install depdendencies for the docs package (using pnpm filters)
2. Builds whatever is necessary (nuqs package), using Turborepo dependencies and remote cache
3. Runs a type-checking script on the docs package (to be added to `packages/docs/package.json`)
4. Runs the build step of the docs package

Note: fumadocs needs a post-install script to generate some types from the MDX content.

This step should run after (ie: depend on) CI core, so that Turborepo can use the cached build output of the nuqs package.

Amirmohammad-Bashiri:
I've opened a pull request that implements the CI job for type-checking and building the docs package as described in this issue.
Please let me know if it has any issues or needs further changes.

PR #1087, opened 2025-08-22
ci: add docs typecheck & build job (#1075)

Amirmohammad-Bashiri:
- Add typecheck script to docs package.json
- Create CI job for docs typecheck and build with Turborepo cache
- Include fumadocs postinstall for MDX types

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -14,6 +14,42 @@ env:
   FORCE_COLOR: 3 # Diplay chalk colors
 
 jobs:
+  docs:
+    name: Docs Typecheck & Build
+    runs-on: ubuntu-24.04-arm
+    needs: [ci-core]
+    steps:
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8
+      - uses: pnpm/action-setup@a7487c7e89a18df4991f7f222e4898a00d66ddda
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-cd.yml
```diff
@@ -14,6 +14,42 @@ env:
...
+      - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020
+        with:
+          node-version-file: .node-version
+          cache: pnpm
+      - name: Install docs dependencies
+        run: pnpm install --frozen-lockfile --filter docs...
+      - name: Run fumadocs postinstall
+        run: pnpm --filter docs run postinstall
+      - name: Build dependencies (nuqs, res packages)
+        run: pnpm build --filter docs... --filter=!docs
```
franky47: **note**: You can use `docs^...` to only target dependencies of the docs package: https://pnpm.io/filtering

franky47 (changes requested):
Thanks, this should work but we could probably simplify it.

`pnpm run build --filter docs...` will build nuqs (which should have been cached as the job depends on CI core) & the docs themselves, and uses the Turborepo cache too.

So we could have:
1. Install deps
2. Type checking
3. Build docs and its dependencies

Could you also move this step down in the file to be under the CI core please? So that the hierarchy is easier to visualise.

Amirmohammad-Bashiri:
Thanks for the feedback.
I’ve updated the CI workflow as suggested
Let me know if anything else needs adjustment.

PR #1095, opened 2025-08-26
chore: Add bundle size comparison workflow

miinhho:
## Summary
This PR introduces an automated bundle size comparison workflow that runs on pull requests affecting the nuqs package. The workflow compares bundle sizes between the base branch and PR head, providing insights into how code changes impact the final bundle size.

## What's Changed
- Added `.github/workflows/size-comparison.yml` workflow
- The workflow automatically runs when PRs modify TypeScript files in `packages/nuqs/src/` or related configuration files
- Uses the existing `size-limit` setup already configured in the `nuqs` package
- Leverages `andresz1/size-limit-action` to generate size comparison reports

## Features
- **Smart triggering**: Only runs when relevant files are changed (TypeScript source files, package.json, or build configs)
- **Test-only optimization**: Skips bundle size checks when only test files are modified to reduce unnecessary CI runs
- **Automated reporting**: Posts size comparison comments directly on PRs via the size-limit-action
- **Incremental updates**: Re-runs comparison when new commits are pushed to the PR
- **Draft PR handling**: Skips execution for draft PRs to conserve resources

image

## Workflow Behavior
1. **Triggers on**: PR opened/synchronized/reopened with changes to nuqs package files
2. **Builds**: Installs dependencies and runs size-limit analysis
3. **Reports**: Automatically comments on PR with size changes (bytes and percentage)
4. **Updates**: Re-evaluates on subsequent commits

## Benefits
- Provides immediate visibility into bundle size impacts of code changes
- Helps prevent accidental bundle size regressions
- Enables informed decisions about performance trade-offs
- Reduces manual overhead of size monitoring

## Testing
- Tested on forked repository with various PR scenarios
- Confirmed proper triggering conditions and size reporting
- Verified test-only file detection works correctly

This workflow will help maintain awareness of bundle size changes and support performance-conscious development practices.

### DIFF
--- a/.github/workflows/size-comparison.yml
+++ b/.github/workflows/size-comparison.yml
@@ -21,7 +21,11 @@ jobs:
       CI_JOB_NUMBER: 1
     steps:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8
-
+      - uses: pnpm/action-setup@a7487c7e89a18df4991f7f222e4898a00d66ddda
+      - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020
+        with:
+          node-version-file: .node-version
+          cache: pnpm
       - name: Check for test file changes only
         id: check-files
         run: |
@@ -37,10 +41,13 @@ jobs:
           done
 
           echo "test_only=$test_only" >> $GITHUB_OUTPUT
-
+      - name: Install dependencies
+        if: steps.check-files.outputs.test_only != 'true'
+        run: pnpm install --ignore-scripts --frozen-lockfile --filter nuqs...
       - name: Check bundle size
         if: steps.check-files.outputs.test_only != 'true'
         uses: andresz1/size-limit-action@v1.8.0
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
+          build_script: pnpm build --filter nuqs
           directory: packages/nuqs
--- a/.github/workflows/size-comparison.yml
+++ b/.github/workflows/size-comparison.yml
@@ -23,6 +23,18 @@ jobs:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8
         with:
           fetch-depth: 0
+      - name: Ensure cache paths exist
+        run: |
+          mkdir -p node_modules
+          mkdir -p .pnpm-store
+      - uses: actions/cache@v3
+        with:
+          path: |
+            node_modules
+            .pnpm-store
+          key: ${{ runner.os }}-pnpm-${{ hashFiles('**/pnpm-lock.yaml') }}
+          restore-keys: |
+            ${{ runner.os }}-pnpm-
       - uses: pnpm/action-setup@a7487c7e89a18df4991f7f222e4898a00d66ddda
       - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020
         with:
@@ -32,8 +44,10 @@ jobs:
         id: check-files
         run: |
           set -e
-          # Fetch full history (if not already)
-          git fetch --unshallow || true
+          # Only unshallow if the repo is shallow
+          if git rev-parse --is-shallow-repository | grep -q true; then
+            git fetch --unshallow
+          fi
 
           # Get changed files between base and head
           changed_files=$(git diff --name-only "${{ github.event.pull_request.base.sha }}" "${{ github.sha }}" -- 'packages/nuqs/src/**/*.ts')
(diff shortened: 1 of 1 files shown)


.github/workflows/size-comparison.yml
```diff
@@ -0,0 +1,49 @@
+name: Bundle Size Comparison
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+    paths:
+      - 'packages/nuqs/src/**/*.ts'
```
franky47, line 6: TIL you could specify paths to watch, that's super useful, thanks!

.github/workflows/size-comparison.yml
```diff
@@ -0,0 +1,49 @@
...
+permissions:
+  pull-requests: write
+
+jobs:
+  size-comparision:
+    name: Compare bundle size
+    runs-on: ubuntu-24.04-arm
+    if: github.event.pull_request.draft == false
+    env:
+      CI_JOB_NUMBER: 1
```
franky47: **question**: What is the purpose of this `CI_JOB_NUMBER: 1` line?

.github/workflows/size-comparison.yml
```diff
@@ -0,0 +1,49 @@
...
+    steps:
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8
+      - name: Ensure cache paths exist
+        run: |
+          mkdir -p node_modules
+          mkdir -p .pnpm-store
+      - uses: actions/cache@v3
```
franky47: **nitpick**: Could you pin `actions/cache` to the latest release's SHA-1 please?

.github/workflows/size-comparison.yml
```diff
@@ -0,0 +1,49 @@
...
+            ${{ runner.os }}-pnpm-
+      - uses: pnpm/action-setup@a7487c7e89a18df4991f7f222e4898a00d66ddda
+      - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020
+        with:
+          node-version-file: .node-version
+          cache: pnpm
+      - name: Install dependencies
+        run: pnpm install --ignore-scripts --frozen-lockfile --filter nuqs...
+      - name: Check bundle size
+        uses: andresz1/size-limit-action@v1.8.0
```
franky47: **nitpick**: Same here, actions should always be pinned to a specific commit (as tags can be moved).

https://francoisbest.com/posts/2020/the-security-of-github-actions

.github/workflows/size-comparison.yml
miinhho: Thank you for pointing that out! I initially included it because it was in the size-limit-action README example. After your question, I checked the size-limit changelog and found that version 0.3.1 mentions using ci-job-number to identify CI job numbers. However, since there's no clear documentation about this environment variable and I've confirmed that the workflow works fine without it, I'll remove it.

.github/workflows/size-comparison.yml
miinhho: Thank you for the feedback! And thanks for sharing helpful article 😄 
Thanks to you, I learned about new security best practices!

miinhho:
> Thanks, could you pin the actions dependencies please?
> 
> What's the cache step (before setting up PNPM & Node.js) doing? I don't have that one in the other workflows, but to be fair the cache behaviour there hasn't been the most stable.

I initially added the cache step before setting up PNPM & Node.js because I was getting errors, but after your feedback, I removed it and configured it the same way as other workflows - and it works perfectly now! Thank you 😄

franky47:
Hum one issue is that we run size-limit as a postbuild script already, so if we blow up the budget, the build step crashes and the action can't pick it up.

I'll see how to shuffle things around.

franky47:
The action errors with "Error creating comment. This can happen for PR's originating from a fork without write permissions.", however we already have the permission `pull-requests: write` 🤔

franky47:
Honestly since the action doesn't seem super well maintained, we could just fork it 😈 (and make it a prettier output while we're at it)

Especially that now we can run TypeScript without a build step in actions 🙌:
https://bsky.app/profile/danielroe.dev/post/3lxczp5apec2x

miinhho:
I found https://github.com/preactjs/compressed-size-action, which made and maintained by preact team.
Here is the pr that used compressed-size-action. https://github.com/toss/suspensive/pull/1760#issuecomment-3340490547

How about using this action?

franky47:
Yeah we could give that one a try, it could also be its own workflow to allow using the [`on.pull_request.paths`](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#onpushpull_requestpaths) filter (as this won't need to run on, say, documentation PRs).

However, it seems to analyse on a file-by-file basis, we'd want to bundle from the entrypoints (dist/index.js and dist/server.js), because the chunks might change hash as updates occur.

PR #1100, opened 2025-08-29
doc: add UUID parser to the shadcn registry

max-programming:
This PR adds a UUID parser validation with 1 to 8 versions support

It can be used like this

```ts
const [id, setId] = useQueryState('id', parseAsUuid())
// or
const [id, setId] = useQueryState('id', parseAsUuid({ version: 7 }))
```


max-programming:
Let me know if something is missing or something needs to modified

max-programming:
@franky47 what might be the reason for the failing build?

franky47:
It's likely `size-limit` that's set too close to the current build size (5.48kB build for a 5.5kB size limit). I can bump it to 6kB to also fit #1036.

packages/nuqs/src/parsers.ts
```diff
@@ -247,6 +247,56 @@ export const parseAsIsoDate: ParserBuilder<Date> = createParser({
...
+ *   'sessionId',
+ *   parseAsUuid({ version: 4 }).withDefault('00000000-0000-0000-0000-000000000000')
+ * )
+ *
+ * // URL: ?sessionId=f47ac10b-58cc-4372-a567-0e02b2c3d479
+ * console.log(sessionId) // "f47ac10b-58cc-4372-a567-0e02b2c3d479"
+ * ```
+ */
+export function parseAsUuid(opts?: {
+  version?: number
```
franky47: **suggestion**: Could we use a number literal here? So we can't pass invalid UUID versions, like 42.
```suggestion
  version?: 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8
```

franky47:
Hey, coming back to this PR, I'm considering adding support in the nuqs docs for the [shadcn CLI](https://ui.shadcn.com/docs/registry/getting-started), to install custom parsers with a one-line command.

Would you mind if we moved this parser to the community section, and have it installable this way?

I'd like to keep the core bundle size small, and this could be one way.

max-programming:
Alright. So what will be the way to handle it? As in where should this parser be moved to?

franky47:
We need to implement the shadcn registry first, see https://github.com/47ng/nuqs/pull/1036#issuecomment-3247952135

max-programming:
Okay! Meanwhile I think I should turn this PR into a draft one

max-programming:
@franky47 now that the shadcn registry work has landed, here's how this parser can be used

```sh
npx shadcn@latest add @nuqs/parser-uuid
```


it follows the same `packages/docs/src/registry/items` layout you pointed `parseAsColor` at in #1342

image
image
image

franky47 (approved):
LGTM, thanks!

PR #1104, opened 2025-08-30
feat(debounce): add warning when shallow: true is used with debounce

I-3B:
show a warning in case `shallow: true` is used with debounce, even with [the clarification](https://nuqs.47ng.com/docs/options#debounce) in docs, users may still trip up and combine the two options.

I've started this PR as a draft because there're some points that can be debated:
- Is a `console.warn` enough or would it be better to have something more severe like `console.error` or even an actual runtime error?
- The options check is currently in `useQueryStates` hook, as this is the only place where `shallow: true` will work (?), but are there any possible places to also include the check in?

### DIFF
--- a/packages/nuqs/src/useQueryStates.ts
+++ b/packages/nuqs/src/useQueryStates.ts
@@ -119,6 +119,11 @@ export function useQueryStates<KeyMap extends UseQueryStatesKeysMap>(
     internalState,
     initialSearchParams
   )
+  if (limitUrlUpdates?.method === 'debounce' && shallow === true) {
+    console.warn(
+      '[nuqs] `shallow: true` and `limitUrlUpdates: debounce` are not compatible, use `shallow: false` instead so debounce can work properly (https://nuqs.47ng.com/docs/options#debounce)'
+    )
+  }
   // Initialise the refs with the initial values
   if (
     Object.keys(queryRef.current).join('&') !==


franky47:
Thanks!

I'd move this warning to the state updater function, to avoid printing it repeatedly on render.

Having it in `useQueryStates` only is a good choice, since `useQueryState` now uses `useQueryStates` internally (so it covers both cases).

One potential nitpick is we could move the error text to the errors.ts file, and assign a status code to it (I had been trying to follow HTTP status codes for similar meaning as a running easter egg, but it's limiting as there isn't a great choice for this case. Maybe 424 Failed Dependency? 🤷‍♂️).

The advantage of a status code is to link to an error document (eg: errors/NUQS-424.md) where we can add more context about why the warning occurs, and remediation steps for users.

I-3B:
> One potential nitpick is we could move the error text to the errors.ts file, and assign a status code to it (I had been trying to follow HTTP status codes for similar meaning as a running easter egg, but it's limiting as there isn't a great choice for this case. Maybe 424 Failed Dependency? 🤷‍♂️).
> 
> The advantage of a status code is to link to an error document (eg: errors/NUQS-424.md) where we can add more context about why the warning occurs, and remediation steps for users.

what about [422 Unprocessable Content](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/422), felt it's more suitable than 424

franky47:
Sounds good, let me fix the size-limit threshold in the main branch so all those PRs have any chance of passing.

packages/nuqs/src/lib/errors.ts
```diff
@@ -3,6 +3,7 @@ export const errors = {
   404: 'nuqs requires an adapter to work with your framework.',
   409: 'Multiple versions of the library are loaded. This may lead to unexpected behavior. Currently using `%s`, but `%s` (via the %s adapter) was about to load on top.',
   414: 'Max safe URL length exceeded. Some browsers may not be able to accept this URL. Consider limiting the amount of state stored in the URL.',
+  422: 'Invalid options combination. `shallow: true` and `limitUrlUpdates: debounce` are not compatible.',
```
franky47: **suggestion**: Since in a lot of cases `shallow: true` is implicit (via defaults), the message should focus on `limitUrlUpdates` first, then mention it should be used with `shallow: false`.

```suggestion
  422: 'Invalid options combination: `limitUrlUpdates: debounce` should be used in SSR scenarios, with `shallow: false`',
```

errors/NUQS-422.md
```diff
@@ -0,0 +1,9 @@
+# Invalid Options Combination.
+
+This error will show up if you combine `shallow: true` and `limitUrlUpdates: debounce` options.
```
franky47: **suggestion**: Wording
```suggestion
This warning will show up if you combine `shallow: true` (the default) and `limitUrlUpdates: debounce` options.
```

errors/NUQS-422.md
```diff
@@ -0,0 +1,9 @@
...
+
+Debounce only makes sense for server-side data fetching, so combining it with `shallow: true` will not work as expected.
```
franky47: **note**: We could add a sentence here explaining that the returned state is always updated instantly.

Even if it's written in the docs, the fewer clicks to information, the better.

franky47 (changes requested):
Thanks, there are a few areas where the wording could be more explicit about what the actual behaviour is, and when debounce makes sense.

PR #1110, opened 2025-09-01
feat: Add processUrlSearchParams middleware as adapter prop

### LINKED ISSUE #1109: feat: Callback to process merged URLSearchParams on update
Let's move the discussion from #1066 here.

The idea is to introduce a callback as an adapter prop that will be called **after** the URLSearchParams have been merged:
https://github.com/47ng/nuqs/blob/97390a55f4d219f9c59367aab094c33ff7888254/packages/nuqs/src/lib/queues/throttle.ts#L177-L183

It should run **before** the URLSearchParams object is sent to the adapter for updating the URL.

This can be used to implement a few interesting behaviours:
- Sorting the search params in some way (eg: alphabetically by key)
- Injecting / removing values
- Other kinds of processing

Example API:

```tsx
<NuqsAdapter 
  processUrlSearchParams={(search) => {
    const entries = Array.from(params.entries())
    entries.sort(([a], [b]) => a.localeCompare(b))
    return new URLSearchParams(entries)
  }}
>
  { children }
</NuqsAdapter>
```

## Message history

> (like a callback function to process URLSearchParams after applying updates for alphabetical sorting).

Is this feature planned? We're definitely looking for this feature specifically. (alphabetical sorting of parameters)
Edit: or a callback where we can do it ourselves.

_Originally posted by @Multiply in https://github.com/47ng/nuqs/issues/1066#issuecomment-3241556409_

---

> @Multiply would you like to try your hand at a PR?

Was already looking into the recent introduction of global default config.
Do you have a preference for where the middleware should be called?

_Originally posted by @Multiply in https://github.com/47ng/nuqs/issues/1066#issuecomment-3241903381_

## Related discussions

- https://github.com/47ng/nuqs/discussions/634
- https://github.com/47ng/nuqs/discussions/777
- https://github.com/47ng/nuqs/discussions/1053

Multiply:
- [x] Add tests
- [x] Add docs
- [x] Add test to verify that useQueryStates picks up a callback change

Closes #1109.

### DIFF
--- a/packages/docs/content/docs/options.client.tsx
+++ b/packages/docs/content/docs/options.client.tsx
@@ -1,6 +1,8 @@
 'use client'
 
 import { Button } from '@/src/components/ui/button'
+import { Checkbox } from '@/src/components/ui/checkbox'
+import { Label } from '@/src/components/ui/label'
 import { parseAsInteger, useQueryState } from 'nuqs'
 import { NuqsAdapter } from 'nuqs/adapters/react'
 import { useEffect, useState } from 'react'
@@ -15,8 +17,18 @@ export function DemoSkeleton() {
   )
 }
 
+function sortAlphabetically(search: URLSearchParams): URLSearchParams {
+  const entries = Array.from(search.entries())
+  entries.sort(([a], [b]) => a.localeCompare(b))
+  return new URLSearchParams(entries)
+}
+function passThrough(search: URLSearchParams): URLSearchParams {
+  return search
+}
+
 export function AlphabeticalSortDemo() {
   const [hydrated, setHydrated] = useState(false)
+  const [enableSorting, setEnableSorting] = useState(true)
   useEffect(() => {
     setHydrated(true)
   }, [])
@@ -27,17 +39,22 @@ export function AlphabeticalSortDemo() {
 
   return (
     <NuqsAdapter
-      processUrlSearchParams={(search) => {
-        const entries = Array.from(search.entries())
-        entries.sort(([a], [b]) => a.localeCompare(b))
-        return new URLSearchParams(entries)
-      }}
+      processUrlSearchParams={enableSorting ? sortAlphabetically : passThrough}
     >
-      <figure className="flex flex-wrap justify-around gap-2 rounded-md border border-dashed p-2">
-        <ComponentToggle id="a" />
-        <ComponentToggle id="b" />
-        <ComponentToggle id="c" />
-      </figure>
+      <>
+        <Label className="flex items-center gap-2">
+          <Checkbox
+            checked={enableSorting}
+            onCheckedChange={checked => setEnableSorting(checked === true)}
+          />{' '}
+          Enable alphabetical sorting on updates
+        </Label>
+        <figure className="not-prose mt-4 mb-8 flex flex-wrap justify-around gap-2 rounded-md border border-dashed p-2">
+          <ComponentToggle id="a" />
+          <ComponentToggle id="b" />
+          <ComponentToggle id="c" />
+        </figure>
+      </>
     </NuqsAdapter>
   )
 }
@@ -54,7 +71,7 @@ export function TimestampDemo() {
 
   return (
     <NuqsAdapter
-      processUrlSearchParams={(search) => {
+      processUrlSearchParams={search => {
         const params = new URLSearchParams(search)
         params.set('ts', Date.now().toString())
         return params
@@ -90,7 +107,7 @@ function ComponentToggle({ id }: { id: string }) {
   return (
     <div className="rounded-xl p-1.5">
       <Button
-        onClick={() => setCount(c => c ? 0 : 1)}
+        onClick={() => setCount(c => (c ? 0 : 1))}
         className="min-w-42 tabular-nums"
       >
         Toggle "{id}"
--- a/packages/docs/content/docs/options.client.tsx
+++ b/packages/docs/content/docs/options.client.tsx
@@ -1,0 +1,110 @@
+'use client'
+
+import { Button } from '@/src/components/ui/button'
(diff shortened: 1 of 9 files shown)


franky47:
There are some linting errors, is your Prettier not set to format on save?

Multiply:
I generally don't use prettier in my own projects, but I've run it now. 🚀

franky47 (approved):
LGTM, thanks! I'll add a section in the docs (we could also have a demo component for this 👀)

Multiply:
> LGTM, thanks! I'll add a section in the docs (we could also have a demo component for this 👀)

If you can give some pointers, I can add it when I get home, later tonight.

franky47:
I was thinking of reusing the same technique I used in the nuqs-2.5 blog post (for the key isolation demo): wrapping a part of the tree with the React SPA adapter, to scope the effect to search params updates happening in its children.

Location-wise, it would probably be best to add a new heading to the options page ("Adapter props"), have the defaultOptions live in there too, and showcase the alphabetical sorting as both a code block for copy-paste and a demo.

The demo could have three buttons updating counters in keys "a", "b", and "c", with a toggle to enable/disable the sorting (ie: switching the `processUrlSearchParams` between actual sorting and a no-op). Which makes me realise the state updater function's `useCallback` might need to have `processUrlSearchParams` in its dependency array for proper tracking.

Multiply:
@franky47 most likely not exactly as you requested, but I hope it's a good starting point.

packages/docs/content/docs/options.mdx
```diff
@@ -394,3 +374,83 @@ const dateParser = createParser({
   eq: (a: Date, b: Date) => a.getTime() === b.getTime() // [!code highlight]
 })
 ```
+
+## Adapter props
+
+### Defaults options
```
franky47: **nitpick**: Can we keep the same title, so that the slug anchor doesn't break?

```suggestion
### Global defaults override
```

packages/docs/content/docs/options.client.tsx
```diff
@@ -0,0 +1,110 @@
...
+    setHydrated(true)
+  }, [])
+
+  if (!hydrated) {
+    return <DemoSkeleton />
+  }
+
+  return (
+    <NuqsAdapter
+      processUrlSearchParams={(search) => {
```
franky47: Ooh this is a neat idea! Maybe not for a timestamp, but adding a version number in the URL would be very handy for migrations. 👀

franky47 (approved):
Looks good, thanks! I'll do some minor wording tweaks and this should be good to merge.

packages/docs/content/docs/options.mdx
Multiply: I didn't think of existing anchors, all good. 👍

Multiply:
> Looks good, thanks! I'll do some minor wording tweaks and this should be good to merge.

Let me know if you need anything else. Happy to test the final changes in our project.
Thanks. :)

franky47:
Feel free to try the pkg.pr.new link in your app, I added reactivity for a change in the callback in useQueryStates, just need to find a clean way to test that.

Multiply:
> Feel free to try the pkg.pr.new link in your app, I added reactivity for a change in the callback in useQueryStates, just need to find a clean way to test that.

I don't have a need for the reactivity bit, but it definitely works as is, if it's static at least.

packages/docs/content/docs/options.client.tsx
Multiply: We've previously used something similar for our preview environments for normal route changes where we just needed a unique timestamp to ensure our CDN didn't serve a cached page.

I couldn't think of other neat situations other than this, and versioning as well.
I didn't go with the latter, as I didn't want to maintain state outside the NuqsAdapter if possible.

franky47:
Thanks for your help! I have a few things I need to work on before this lands in 2.6.0 (likely by the end of the week).

packages/docs/content/docs/options.client.tsx
franky47: Versioning (and by extension, migrations) is something I want to work on once some of the major roadmap items have been ironed out.

https://nuqs.47ng.com/blog/beware-the-url-type-safety-iceberg#immutability

packages/docs/content/docs/options.mdx
```diff
@@ -394,3 +372,91 @@ const dateParser = createParser({
...
+Sort the search parameters alphabetically:
+
+```tsx
+// [!code word:processUrlSearchParams]
+<NuqsAdapter
+  processUrlSearchParams={(search) => {
+    const entries = Array.from(search.entries())
+    entries.sort(([a], [b]) => a.localeCompare(b))
+    return new URLSearchParams(entries)
+  }}
```
hugotiger, line 425: Was it considered to recommend the usage of the native [`sort()`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/sort) function on `URLSearchParams`?

Something along the lines of:
```suggestion
  processUrlSearchParams={(search) => {
    const clone = new URLSearchParams(search)
    clone.sort()
    return clone
  }}
```
Whilst it isn't the exact same sorting algorithm, I was thinking maybe it'd be preferable to recommend following the standard by default!

I can update the docs if you agree 😄

Multiply: If we don't care about immutability, simply call `search.sort()` and returning `search` directly could also do.

franky47: Imagine building an entire library around URLSearchParams and only discovering the .sort function now 😅

Yes standards would be better to advertise in the docs. It's a shame it doesn't return itself though, that could have been a one-liner.

Multiply: https://github.com/47ng/nuqs/pull/1113

PR #1113, opened 2025-09-02
doc: use URLSearchParams.sort

Multiply:


franky47:
**suggestion**: A few things we could improve:

- Add a [QuerySpy](https://github.com/47ng/nuqs/blob/0f68f93461a7a3e2e3dd4142daf4c40f2b1d15fd/packages/docs/src/components/query-spy.tsx) to the demos, that only showcases that demo's URL keys (using the `keepKeys` prop). That'll help folks on mobile or in browsers where the URL is hidden by default (heresy!)
- Move the note about copy vs in-place to the alphabetical sorting example (first occurrence)

Multiply:
> **suggestion**: A few things we could improve:
> 
> * Add a [QuerySpy](https://github.com/47ng/nuqs/blob/0f68f93461a7a3e2e3dd4142daf4c40f2b1d15fd/packages/docs/src/components/query-spy.tsx) to the demos, that only showcases that demo's URL keys (using the `keepKeys` prop). That'll help folks on mobile or in browsers where the URL is hidden by default (heresy!)
> * Move the note about copy vs in-place to the alphabetical sorting example (first occurrence)

I can't seem to get the `<QuerySpy>` to work outside of initial page load. It simply doesn't react to any state changes, regardless of it being used inside/outside of the `<NuqsAdapter>`.

franky47:
Ah yeah good point, it uses the Next.js APIs, but the demo is scoped in the React SPA. It's odd as Next.js usually picks up history changes made from other sources. I'll have a look later.

franky47:
I found the issue: Next.js only listens to external history API calls when the state passed is `null`, but the React SPA adapter maintains whatever it was before. I'll see if I can add an option in the React SPA adapter to control this behaviour.

Edit: actually, using the Next.js app router adapter (nested into itself) also works, which is fine for this case (in the nuqs 2.5 blog post it was to showcase key isolation, which behaves differently in both, but the API is the same on all adapters here).

PR #1115, opened 2025-09-03
feat: add processUrlSearchParams to createSerializer

Multiply:


Multiply:
@franky47 I realized I needed this for our SEO-friendly/canonical URLs as well. Let me know if you want it implemented differently. :)

packages/docs/content/docs/utilities.mdx
```diff
@@ -99,6 +99,33 @@ serialize({
...
+
+```ts
+// [!code word:processUrlSearchParams]
+const serialize = createSerializer(
+  {
+    foo: parseAsString,
+  },
+  {
+    processUrlSearchParams: (search) => {
+      search.set('processed', 'true')
```
franky47: **nitpick**: Showcasing the sorting here (with a mention about SEO) might make more sense than injecting values (which can already be done with the base argument).

Multiply: I can reuse the original example and keep it sorted, does that sound good?

Multiply: I ended up using something a bit more concise and also added a similar test for consistency.

Feel free to do more changes.

PR #1117, opened 2025-09-04
doc: Use SVG for Next.js logo in the adapters ToC

franky47:
This showcases a bug with Chrome & Firefox where the gradient on the N logo
disappears when the ToC item is active, but shows up again when it's inactive.

It behaves correctly in Safari.

### DIFF
--- a/packages/docs/content/docs/adapters.mdx
+++ b/packages/docs/content/docs/adapters.mdx
@@ -24,9 +24,7 @@ wrapping it with a `NuqsAdapter{:ts}` context provider:
 - <ReactRouterV7 className='inline mr-1.5' role="presentation" /> [React Router v7](#react-router-v7)
 - <TanStackRouter className='inline mr-1.5 not-prose' role="presentation"/> [TanStack Router](#tanstack-router)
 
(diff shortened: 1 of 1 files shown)


franky47:
Note to self: the tip from @fuma-nama was to move the NextJS SVG logo component to a client component, so that the `useId` hook actually separates gradient definitions. And it works! 🤯

From the [Next.js Discord](https://discord.com/channels/752553802359505017/1159510933647925248/1413091158070001735):

CleanShot 2025-09-06 at 22 18 49@2x


If used as a server component, within the MDX body the IDs are different, but the one in the ToC shares the ID of the first occurrence in the MDX body, and this is what seems to cause issues.

Now the trick is to determine which is best in terms of transfer size: image vs client bundle.

franky47:
Here are the results (loading the adapters doc page from a clean cache): client component wins, at the expense of needing a hydration step. I hate this so much..

```
With use client SVG: 
55 requests | 3,75 MB / 923,89 kB transferred

With image:
57 requests | 3,77 MB / 935,25 kB transferred

With server sprite:
55 requests | 3,76 MB / 930,07 kB transferred
```

PR #1123, opened 2025-09-06
doc: Add sitemap

franky47:
- [x] Add actual file modification date to sitemap
- [ ] Find ways to iterate over the rest of the pages

Maybe drop all of this and use https://www.npmjs.com/package/next-sitemap

A validator that doesn't suck: https://ismywebsiteready.com/tools/sitemap-xml-validator

### DIFF
--- a/packages/docs/src/lib/url.ts
+++ b/packages/docs/src/lib/url.ts
@@ -1,0 +1,7 @@
+export function getBaseUrl() {
+  return process.env.VERCEL_ENV === 'production'
+    ? 'https://' + process.env.VERCEL_PROJECT_PRODUCTION_URL
+    : process.env.VERCEL_URL
+      ? 'https://' + process.env.VERCEL_URL
+      : ''
+}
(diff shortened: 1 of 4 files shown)


packages/docs/src/lib/url.ts
```diff
@@ -0,0 +1,7 @@
+export function getBaseUrl() {
+  return process.env.VERCEL_ENV === 'production'
+    ? 'https://' + process.env.VERCEL_PROJECT_PRODUCTION_URL
+    : process.env.VERCEL_URL
+      ? 'https://' + process.env.VERCEL_URL
+      : ''
```
franky47, line 6: **note**: This doesn't handle "preproduction" deployment URLs well.

Eg, for `next.nuqs.dev` (connected to the `next` Git branch), it still returns the random-based preview deployment ID, where it should use the domain associated with it.

I had this issue with francoisbest.com, the key was to bypass the deployment promotion system IIRC.

PR #1124, opened 2025-09-07
doc: Add contributors section to landing page with GitHub API integration

### LINKED ISSUE #1122: Add list of contributors on the landing page
We have a section for Sponsors & "Used by", but listing the contributors on the landing page would be nice.

- Filter out the bots
- Order by number of contributions
- Avatar grid (like the "Used by" leaderboard, but maybe bigger avatars), with clickable links to each contributor's profile
- Fetched from the GitHub API, and cached for 24h (ISR)
- A merged PR could invalidate the contributors' ISR cache, the same way we invalidate the NPM cache on release and the GHA cache on testing Next.js versions

Agastya18:
## Summary
Adds a new contributors section to the landing page that displays GitHub contributors in an avatar grid, positioned between the Sponsors and "Used by" sections.
fix: #1122 

## Features
- **GitHub API integration**: Fetches contributors from `47ng/nuqs` repository
- **Bot filtering**: Automatically filters out bots and bot accounts
- **Sorted by contributions**: Orders contributors by contribution count (descending)
- **Responsive avatar grid**: Larger avatars  section with hover effects
- **Clickable profiles**: Each avatar links to the contributor's GitHub profile
- **24h ISR caching**: Uses Next.js ISR with `contributors` tag for performance
- **Rate limit friendly**: Optional `GITHUB_TOKEN` for higher API limits

## Technical Details
- New component: `packages/docs/src/app/(pages)/_landing/contributors.tsx`
- Added to landing page: `packages/docs/src/app/(pages)/page.tsx`
- ISR revalidation: Added `contributors` tag to existing ISR endpoint
- Auto-refresh: Contributors cache invalidates on release via GitHub Actions

## SS
Screenshot 2025-09-08 at 12 04 01 AM

Closes #1122.


.github/workflows/ci-cd.yml
```diff
@@ -405,3 +405,7 @@ jobs:
         run: |
           curl -s "https://nuqs.47ng.com/api/isr?tag=npm&token=${{ secrets.ISR_TOKEN }}"
           curl -s "https://nuqs.dev/api/isr?tag=npm&token=${{ secrets.ISR_TOKEN }}"
+      - name: Invalidate ISR cache for contributors in the docs
```
franky47, line 408: **note**: This step should not run in CI, but rather when a PR is merged. CI runs when commits are pushed to a PR or on named branches (next, master, beta etc).

Maybe an if guard here that only triggers on pushes on next might make sense though, since merged branches result in a commit there.

packages/docs/src/app/(pages)/_landing/contributors.tsx
```diff
@@ -0,0 +1,92 @@
...
+  html_url: z.url(),
+  avatar_url: z.url(),
+  type: z.string(),
+  contributions: z.number()
+})
+type Contributor = z.infer<typeof contributorSchema>
+
+async function fetchContributors(): Promise<Contributor[]> {
+  const url = new URL('https://api.github.com/repos/47ng/nuqs/contributors')
+  url.searchParams.set('per_page', '27')
```
franky47: **question**: 27 is oddly specific, why that number particularly?

packages/docs/src/app/(pages)/_landing/contributors.tsx
```diff
@@ -0,0 +1,92 @@
...
+  }
+  const data = await res.json()
+  const contributors = z.array(contributorSchema).parse(data)
+
+  // Filter bots (type Bot, or login including [bot], or known bot accounts)
+  const isHuman = (c: Contributor) => {
+    const loginLower = c.login.toLowerCase()
+    if (c.type === 'Bot') return false
+    if (loginLower.endsWith('[bot]')) return false
+    if (loginLower.includes('bot')) return false
```
franky47: **warning**: This will falsely flag humans with a name like "Abbot" or "Bottega" as bots.

franky47 (changes requested):
Thanks, the bot detection logic needs a little more work, we could check against a set of known account IDs (dependabot across its many versions, Depfu, Renovate, etc).

If the list is easy enough to edit, it will be trivial to add more in the future when we see more bots creep through the (human) contributors list.

packages/docs/src/app/(pages)/_landing/contributors.tsx
Agastya18: I picked 27 intentionally because it fits the alignment of the contributor section.with 2 rows of avatars it looks clean, if u want i can increase it.

franky47: Ah I see, 27 minus the bots makes 24 which fits nicely on the grid, gotcha.

I'd be more in favour of having all contributors being represented, and use a flex wrap layout for overflows. Might need a bit less gap between the avatars then.

franky47 (approved):

franky47:
Looks good! I like the progressive size of sponsors > contributors > users.

PR #1128, opened 2025-09-10
fix: Use TanStack Router default stringify for search params

### LINKED ISSUE #1127: [bug] TanStack Router adapter: search params with array of objects serialized as [object Object]
## Context

```
"nuqs": "2.6.0"
```

What framework are you using?

- ✅Other (TanStack Router)

```
"@tanstack/react-router": "1.x"
"react": "18.x"
```

## Description

When using Nuqs with the TanStack Router adapter,  
search params containing objects or arrays of objects are incorrectly serialized into strings like: `[object Object]`

This causes parsing errors when using `parseAsJson` or custom parsers (e.g. shadcn-table sorting state).

**Expected behavior:**  
The adapter should follow TanStack Router’s default serialization (`JSON.stringify` / `JSON.parse`) so that objects/arrays are stored as valid JSON strings in the URL.

We previously discussed this issue here:
https://github.com/47ng/nuqs/pull/953#issuecomment-3272900821

Reference docs:  
https://tanstack.com/router/latest/docs/framework/react/guide/custom-search-param-serialization

gensmusic:
# Fix TanStack Router search params serialization

## Summary

This PR fixes #1127 an issue in the Nuqs TanStack Router adapter where search params 
containing objects or arrays were being incorrectly serialized into `[object Object]`.

## Changes

- Replaced the previous manual flattening logic with `stringifySearchWith(JSON.stringify)` from TanStack Router.
- Ensures that nested objects/arrays in search params are serialized as valid JSON strings.
- Keeps the adapter behavior consistent with TanStack Router defaults, reducing differences between routers (e.g., Next.js vs TSR).

## Notes

- This PR is intentionally kept small and focused on fixing the core serialization issue.
- I’m not deeply familiar with Nuqs internals, so additional features or improvements to the TSR adapter can be added in future PRs.
- Related discussion: [PR #953 comment](https://github.com/47ng/nuqs/pull/953#issuecomment-3272900821)
- Issue tracking this bug: [Issue #1127](https://github.com/47ng/nuqs/issues/1127)

Fixes #1127.

### DIFF
--- a/packages/nuqs/src/adapters/tanstack-router.ts
+++ b/packages/nuqs/src/adapters/tanstack-router.ts
@@ -1,8 +1,16 @@
-import { useLocation, useMatches, useNavigate } from '@tanstack/react-router'
+import {
+  stringifySearchWith,
+  useLocation,
+  useMatches,
+  useNavigate
+} from '@tanstack/react-router'
 import { startTransition, useCallback, useMemo } from 'react'
 import { renderQueryString } from '../lib/url-encoding'
 import { createAdapterProvider, type AdapterProvider } from './lib/context'
 import type { AdapterInterface, UpdateUrlFunction } from './lib/defs'
+
+// Use TanStack Router's default JSON-based search param serialization
+const defaultStringifySearch = stringifySearchWith(JSON.stringify)
 
 function useNuqsTanstackRouterAdapter(watchKeys: string[]): AdapterInterface {
   const search = useLocation({
@@ -20,25 +28,10 @@ function useNuqsTanstackRouterAdapter(watchKeys: string[]): AdapterInterface {
   })
   const searchParams = useMemo(
     () =>
-      // search is a Record<string, string | number | object | Array<string | number>>,
-      // so we need to flatten it into a list of key/value pairs,
-      // replicating keys that have multiple values before passing it
-      // to URLSearchParams, otherwise { foo: ['bar', 'baz'] }
-      // ends up as { foo → 'bar,baz' } instead of { foo → 'bar', foo → 'baz' }
-      new URLSearchParams(
-        Object.entries(search).flatMap(([key, value]) => {
-          if (Array.isArray(value)) {
-            return value.map(v => [key, v])
-          } else if (typeof value === 'object' && value !== null) {
-            // TSR JSON.parses objects in the search params,
-            // but parseAsJson expects a JSON string,
-            // so we need to re-stringify it first.
-            return [[key, JSON.stringify(value)]]
-          } else {
-            return [[key, value]]
-          }
-        })
-      ),
+      // Use TSR’s default stringify to convert search object → URLSearchParams.
+      // This avoids issues where arrays/objects were previously flattened
+      // into invalid values like "[object Object]".
+      new URLSearchParams(defaultStringifySearch(search)),
     [search, watchKeys.join(',')]
   )
 
--- a/packages/nuqs/src/adapters/tanstack-router.ts
+++ b/packages/nuqs/src/adapters/tanstack-router.ts
@@ -1,18 +1,42 @@
 import {
   stringifySearchWith,
+  parseSearchWith,
   useLocation,
   useMatches,
-  useNavigate
+  useNavigate,
+  type AnySchema
 } from '@tanstack/react-router'
-import { startTransition, useCallback, useMemo } from 'react'
+import {
+  createContext,
+  createElement,
+  startTransition,
+  useCallback,
+  useContext,
+  useMemo,
+  type ReactElement,
+  type ReactNode
+} from 'react'
 import { renderQueryString } from '../lib/url-encoding'
-import { createAdapterProvider, type AdapterProvider } from './lib/context'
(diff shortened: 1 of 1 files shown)


franky47:
Thanks, that's a better approach indeed.

However, it would not work for folks who have customised their (de)serialisation layer (say, to encode all search params as base64, or using jsurl2):
https://tanstack.com/router/latest/docs/framework/react/guide/custom-search-param-serialization

One way we could support this is to pass the `stringifySearchWith` as an adapter prop, propagate it down via Context, and read it in the adapter hook. See how it's done in the React SPA adapter, there are a few contexts there to pass stuff down.

Now this would be hard to test though, as it's a global config on the router, and the e2e tests are asserting on the default SerDe layer.

For the record, the previous approach wouldn't have worked with a custom SerDe either, so there's that.. 😅

gensmusic:
@franky47 I just add the support for custom stringifySearchWith

gensmusic:
Please ignore the latest commit for now, it's far more complicated than I thought!

gensmusic:
**This change has been tested locally and works as expected.**

# Parsing search params

Inside TanStack Router, `parseSearchWith`(default or custom) has already been applied, so the `search` we get in our code is an object that has already been parsed.

However, nuqs expects `URLSearchParams`. To make it compatible, we use the default `defaultStringifySearch` (based on JSON.stringify) to convert the object back into a format nuqs can handle correctly.

# Synchronizing search params

When nuqs updates the search params, the `UpdateUrlFunction` receives nuqs-generated `URLSearchParams`.

If there is a custom `stringifySearchWith`, We first use the default `defaultParseSearch` to convert  the nuqs-generated `URLSearchParams` into an object, then pass that object through the user-defined `stringifySearchWith` to produce the desired URLSearchParams.

If no custom stringifySearchWith is provided, we simply fall back to the default behavior, which is compatible with nuqs’ expectations.

```text
        ┌─────────────────────┐
        │   Browser URL query  │
        └──────────┬──────────┘
                   │
                   ▼
        ┌─────────────────────┐
        │ TanStack Router      │
        │ parseSearchWith(...) │
        └──────────┬──────────┘
                   │ produces search object
                   ▼
        ┌─────────────────────┐
        │ Our Adapter          │
        │ defaultStringify     │
        │ → URLSearchParams    │
        └──────────┬──────────┘
                   │ passed to nuqs
                   ▼
        ┌─────────────────────┐
        │        nuqs          │
        │ consumes searchParams │
        └──────────┬──────────┘
                   │
                   ▼
        ┌─────────────────────┐
        │ User updates search  │
        │ params in nuqs       │
        └──────────┬──────────┘
                   │ provides URLSearchParams
                   ▼
        ┌─────────────────────┐
        │ Our Adapter          │
        │ defaultParseSearch   │
        │ → object → stringify │
        │ (custom or default)  │
        └──────────┬──────────┘
                   │
                   ▼
        ┌─────────────────────┐
        │ TanStack Router      │
        │ navigate(to=...)     │
        └─────────────────────┘
```

gensmusic:
@franky47 If you have a moment, I’d love your review on this PR.

franky47:
Thanks, I'll give it a try tonight.

packages/nuqs/src/adapters/tanstack-router.ts
```diff
@@ -18,32 +50,33 @@ function useNuqsTanstackRouterAdapter(watchKeys: string[]): AdapterInterface {
...
+  const searchParams = useMemo(() => {
+    // Regardless of whether the user specified a custom parseSearchWith,
+    // the search object here is already the result after parsing.
+    // We use the default defaultStringifySearch to convert the search
+    // to search params that nuqs can handle correctly.
+    //
+    // Use TSR's default stringify to convert search object → URLSearchParams.
+    // This avoids issues where arrays/objects were previously flattened
+    // into invalid values like "[object Object]".
+    return new URLSearchParams(defaultStringifySearch(search))
```
TkDodo, line 62: I looked at this PR in the context of multi-parsers (https://github.com/47ng/nuqs/pull/1134), but they completely stop working in TanStack Router when I apply the changes from this PR.

The problem seems to be this call to `defaultStringifySearch`. For example, if `search` is:

```
{
  test: [1,2]
}
```

The call to `defaultStringifySearch` turns that into `"?test=%5B1%2C2%5D"`, which gives us `searchParams` with a size of one where the value for `test` is the string `[1,2]`.

This wasn’t the case with the previous implementation, where `searchParams` would correctly have a size of two.

franky47: Is there a way in TanStack Router to **write** to multiple keys?

From what I tested, TSR itself will deserialise `?a=1&a=2&a=3&b=[1,2,3]` into:

```js
{
  a: [1, 2, 3],
  b: [1, 2, 3]
}
```

There is a loss of information on read of whether we use a JSON array or a repeat of keys.

When writing with Route.useNavigate (ie: not using nuqs, and assuming we have a validateSearch set to `z.array(z.number())` for both a & b), they both render to a JSON array in the URL.

Issue #1099, opened 2025-08-29
State temporarily resets to previous value on subsequent rerenders


georgwittberger:
## Context

What's your version of `nuqs`?

```
2.5.1
```

What framework are you using?

- ✅ Next.js (app router)
- ❌ Next.js (pages router)
- ❌ React SPA (no router)
- ❌ Remix
- ❌ React Router
- ❌ Other (please specify)

Which version of your framework are you using?

```
15.5.2
```

## Description

After setting a new value the state returned by `useQueryState` temporarily resets to its previous value when the component is rerendered multiple times due to some other state changes (e.g. loading new data based on new state).

After calling the state setter function of `useQueryState` with a new value the component rerenders four times with the newly set value. That's fine. But if this new value causes other state changes (e.g. data fetching) there are two rerenders where `useQueryState` suddenly returns the previous value. Then on further rerenders the newly set value is returned again.

Taking the reproduction demo a `console.log` directly in the rendering process outputs the following (annotated with comments for better explanation):

```text
// Click on button which sets the state "desc"
Rendering with sort order: desc
Rendering with sort order: desc
Rendering with sort order: desc
Rendering with sort order: desc
// Now the unexpected two renders with the previous value !!!
Rendering with sort order: asc
Rendering with sort order: asc
// Now the value switches back to new value again
Rendering with sort order: desc
Rendering with sort order: desc
Rendering with sort order: desc
Rendering with sort order: desc
...
```

## Reproduction

https://github.com/georgwittberger/nuqs-errors-repro

https://stackblitz.com/github/georgwittberger/nuqs-errors-repro

1. Install dependencies, then start dev server: `npm run dev`
2. Open demo page: http://localhost:3000/reset
3. Open browser dev tools and view Console tab.
4. Click button "Sort desc".
5. Notice there are rerenders with the previous value "asc" after new value "desc" has already been set.

This also fails unit tests that assert the new value.

Run `npm run test` and see failing test in `src/app/reset/page.test.tsx`. This test is expected to succeed.

After data for new state "desc" has been rendered the test simply checks if the rendered state value is still "desc" but fails because it hits the render output with the previous value "asc".

Jackardios:
I encountered the same problem on react-router v7.8.2, nuqs v2.6.0

Jackardios:
I found that the return to the previous state occurs in the useEffect hook
[...]

PR #1133, opened 2025-09-12
feat: add memory option in the testing adapter

franky47:
Partial solution for the #1099 issue, where the state transients through an old value during testing.

This is not the most ideal solution as it may need a change in core, or in how the testing adapter resets its queue, but the lack of memory (state updates only using the initial search params as a base for updates) caused me a few headaches in tests, and this can help make the testing setup behave more closely to other adapters.

- [x] Add a note in the docs that this "memory" is per-adapter, and so is isolated between tests, but shared for components under the same adapter.


franky47:
@georgwittberger could you try the following deployment build in your setup please? I confirmed it passes the reset test in your reproduction repository.

```
pnpm add https://pkg.pr.new/nuqs@1133
```

You'll need to add the `hasMemory: true` prop on the `<NuqsTestingAdapter>`, or `withNuqsTestingAdapter` wrapper, to let the testing adapter record state updates internally, like other frameworks would.

Without memory, all state updates are based on the initial search params, which might work for testing units of behaviour, but causes problems with the optimistic state flow you described (brilliantly) in https://github.com/47ng/nuqs/issues/1099#issuecomment-3281092278.

There may be a better fix by seeing how to reset the queue differently in the testing adapter, I'll see if I can rework this PR later (but I think the memory mode has value on its own, for testing multiple updates in a more intuitive way).

georgwittberger:
Yes, this works for the component test. Should solve the issues with the testing adapter.

PR #1134, opened 2025-09-13
feat: multi-parsers

TkDodo:
> **note**: Description by @franky47 

MultiParsers allow **key repetition** in the URL.

E.g.: with the built-in `parseAsNativeArrayOf`:
```ts
const [state] = useQueryState('key', parseAsNativeArrayOf(parseAsInteger))
// url:   /?key=1&key=2&key=3
// state: [1, 2, 3]
```

Creating custom multi-parsers allow reducing the array of query values into more complex objects:
```ts
// /?kv=foo:bar&kv=baz:qux
{
  foo: 'bar',
  baz: 'qux'
}
```

### DIFF
--- a/packages/docs/content/docs/parsers/built-in.mdx
+++ b/packages/docs/content/docs/parsers/built-in.mdx
@@ -329,11 +329,11 @@ parseAsJson(userSchema.validateSync)
 
 If you want to use the native URL format for arrays, repeating the same key multiple times like:
 
-```
-/products?tag=books&tag=tech&tag=design
-```
+import { Querystring } from '@/src/components/querystring'
 
-you can now use `MultiParsers` like `parseAsNativeArrayOf` to read and write those values in a fully type-safe way.
+<Querystring path="/products" value='?tag=books&tag=tech&tag=design' />
+
+you can now use `MultiParsers{:ts}` like `parseAsNativeArrayOf{:ts}` to read and write those values in a fully type-safe way.
 
 ```tsx
 import { useQueryState, parseAsNativeArrayOf, parseAsInteger } from 'nuqs'
@@ -343,15 +343,17 @@ const [projectIds, setProjectIds] = useQueryState(
   parseAsNativeArrayOf(parseAsInteger)
 )
 
-// ?project=123&project=456 ==> [123, 456]
+// ?project=123&project=456 → [123, 456]
 ```
 
 <Suspense fallback={<DemoFallback />}>
   <NativeArrayParserDemo />
 </Suspense>
 
-<Callout title="Empty Array Default">
-  Note that `parseAsNativeArrayOf` has a built-in default value of empty array (`.withDefault([])`) so that you don't have to handle `null` cases.
+<Callout title="Note: empty array default">
+  `parseAsNativeArrayOf{:ts}` has a built-in default value of empty array (`.withDefault([]){:ts}`) so that you don't have to handle `null{:ts}` cases.
+
+  Calls to `.withDefault(){:ts}` can be chained, so you can use it to set a custom default.
 </Callout>
 
 ## Using parsers server-side
--- a/packages/docs/content/docs/parsers/built-in.mdx
+++ b/packages/docs/content/docs/parsers/built-in.mdx
@@ -338,13 +338,12 @@ you can now use `MultiParsers` like `parseAsNativeArrayOf` to read and write tho
 ```tsx
 import { useQueryState, parseAsNativeArrayOf, parseAsInteger } from 'nuqs'
 
-export function ProjectsFilter() {
-  // ?project=123&project=456 ==> [123, 456]
-  const [projectIds, setProjectIds] = useQueryState(
-    'project',
-    parseAsNativeArrayOf(parseAsInteger)
-  )
-}
+const [projectIds, setProjectIds] = useQueryState(
+  'project',
+  parseAsNativeArrayOf(parseAsInteger)
+)
+
+// ?project=123&project=456 ==> [123, 456]
 ```
 
 <Suspense fallback={<DemoFallback />}>
--- a/packages/docs/content/docs/parsers/built-in.mdx
+++ b/packages/docs/content/docs/parsers/built-in.mdx
@@ -325,7 +325,7 @@ parseAsJson(userSchema.validateSync)
 
 ## Native Arrays
 
-<FeatureSupportMatrix introducedInVersion='2.6.0' />
+<FeatureSupportMatrix introducedInVersion='2.7.0' />
 
 If you want to use the native URL format for arrays, repeating the same key multiple times like:
 
--- a/packages/docs/content/docs/parsers/built-in.mdx
+++ b/packages/docs/content/docs/parsers/built-in.mdx
@@ -15,7 +15,8 @@ import {
   DateISOParserDemo,
   DatetimeISOParserDemo,
   DateTimestampParserDemo,
-  JsonParserDemo
+  JsonParserDemo,
+  NativeArrayParserDemo
 } from '@/content/docs/parsers/demos'
 
 Search params are strings by default, but chances are your state is more complex than that.
@@ -320,6 +321,38 @@ parseAsJson(userSchema.validateSync)
 <Callout title="Note">
   Validation functions must either throw an error or
   return `null{:ts}` for invalid data. Only **synchronous** validation is supported.
+</Callout>
+
+## Native Arrays
+
+<FeatureSupportMatrix introducedInVersion='2.6.0' />
+
+If you want to use the native URL format for arrays, repeating the same key multiple times like:
+
+```
+/products?tag=books&tag=tech&tag=design
+```
+
+you can now use `MultiParsers` like `parseAsNativeArrayOf` to read and write those values in a fully type-safe way.
+
+```tsx
+import { useQueryState, parseAsNativeArrayOf, parseAsInteger } from 'nuqs'
+
+export function ProjectsFilter() {
+  // ?project=123&project=456 ==> [123, 456]
+  const [projectIds, setProjectIds] = useQueryState(
+    'project',
+    parseAsNativeArrayOf(parseAsInteger)
+  )
+}
+```
+
+<Suspense fallback={<DemoFallback />}>
+  <NativeArrayParserDemo />
+</Suspense>
+
+<Callout title="Empty Array Default">
+  Note that `parseAsNativeArrayOf` has a built-in default value of empty array (`.withDefault([])`) so that you don't have to handle `null` cases.
 </Callout>
 
 ## Using parsers server-side
--- a/packages/e2e/shared/specs/native-array.tsx
+++ b/packages/e2e/shared/specs/native-array.tsx
@@ -3,7 +3,7 @@
 import { parseAsInteger, parseAsNativeArrayOf, useQueryState } from 'nuqs'
 import { Display } from '../components/display'
 
-export const parser = parseAsNativeArrayOf(parseAsInteger).withDefault([])
+export const parser = parseAsNativeArrayOf(parseAsInteger)
 
 export function NativeArray() {
   const [state, setState] = useQueryState('test', parser)
--- a/packages/e2e/shared/specs/native-array.tsx
+++ b/packages/e2e/shared/specs/native-array.tsx
@@ -1,0 +1,24 @@
+'use client'
+
+import { parseAsInteger, parseAsNativeArrayOf, useQueryState } from 'nuqs'
+import { Display } from '../components/display'
+
+export const parser = parseAsNativeArrayOf(parseAsInteger)
+
+export function NativeArray() {
+  const [state, setState] = useQueryState('test', parser)
+  return (
+    <>
+      <button onClick={() => setState([])}>Reset</button>
+      <button
+        id="add-button"
+        onClick={() =>
+          setState(prev => (prev ?? []).concat((prev ?? []).length + 1))
+        }
+      >
+        Add
+      </button>
+      <Display environment="client" target="name" state={state?.join(' - ')} />
+    </>
+  )
+}
--- a/packages/e2e/shared/specs/native-array.tsx
+++ b/packages/e2e/shared/specs/native-array.tsx
@@ -3,7 +3,7 @@
 import { parseAsInteger, parseAsNativeArrayOf, useQueryState } from 'nuqs'
 import { Display } from '../components/display'
 
-export const parser = parseAsNativeArrayOf(parseAsInteger)
+export const parser = parseAsNativeArrayOf(parseAsInteger).withDefault([])
 
 export function NativeArray() {
   const [state, setState] = useQueryState('test', parser)
@@ -12,9 +12,7 @@ export function NativeArray() {
       <button onClick={() => setState([])}>Reset</button>
       <button
         id="add-button"
-        onClick={() =>
-          setState(prev => (prev ?? []).concat((prev ?? []).length + 1))
-        }
+        onClick={() => setState(prev => prev.concat(prev.length + 1))}
       >
         Add
       </button>
--- a/packages/e2e/tanstack-router/src/routes/native-array.tsx
+++ b/packages/e2e/tanstack-router/src/routes/native-array.tsx
@@ -1,0 +1,13 @@
+import { createFileRoute } from '@tanstack/react-router'
+import { NativeArray, parser } from 'e2e-shared/specs/native-array'
+import { createStandardSchemaV1 } from 'nuqs'
+
+const validateSearch = createStandardSchemaV1(
+  { test: parser },
+  { partialOutput: true }
+)
+
+export const Route = createFileRoute('/native-array')({
+  component: NativeArray,
+  validateSearch
+})
--- a/packages/nuqs/src/lib/compare.ts
+++ b/packages/nuqs/src/lib/compare.ts
@@ -1,6 +1,6 @@
-import type { QueryParam } from './search-params'
+import type { Query } from './search-params'
 
-export function compareQuery<T extends QueryParam>(
+export function compareQuery<T extends Query>(
   a: T | null,
   b: T | null
 ): boolean {
--- a/packages/nuqs/src/lib/compare.ts
+++ b/packages/nuqs/src/lib/compare.ts
@@ -10,7 +10,7 @@ export function compareQuery<T extends string | Iterable<string>>(
   }
   // we expect either strings or iterables, not a mix of both
   if (typeof a === 'string' || typeof b === 'string') {
-    return a === b
+    return false
   }
 
   const iterA = a[Symbol.iterator]()
--- a/packages/nuqs/src/lib/compare.ts
+++ b/packages/nuqs/src/lib/compare.ts
@@ -1,0 +1,33 @@
+export function compareQuery<T extends string | Iterable<string>>(
+  a: T | null,
+  b: T | null
+): boolean {
+  if (a === b) {
+    return true // Referentially stable
+  }
+  if (a === null || b === null) {
+    return false
+  }
+  // we expect either strings or iterables, not a mix of both
+  if (typeof a === 'string' || typeof b === 'string') {
+    return a === b
+  }
+
+  const iterA = a[Symbol.iterator]()
+  const iterB = b[Symbol.iterator]()
+
+  while (true) {
+    const nextA = iterA.next()
+    const nextB = iterB.next()
+
+    if (nextA.done && nextB.done) {
+      return true // both ended at the same time
+    }
+    if (nextA.done !== nextB.done) {
+      return false // different lengths
+    }
+    if (nextA.value !== nextB.value) {
+      return false // mismatched value
+    }
+  }
+}
--- a/packages/nuqs/src/lib/compare.ts
+++ b/packages/nuqs/src/lib/compare.ts
@@ -1,4 +1,6 @@
-export function compareQuery<T extends string | Iterable<string>>(
+import type { QueryParam } from './search-params'
+
+export function compareQuery<T extends QueryParam>(
   a: T | null,
   b: T | null
 ): boolean {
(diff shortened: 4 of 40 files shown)


TkDodo:
`size-limit` fails, and this runs right after build, so before any tests are running or even type-checking is done. I don’t think that’s the best workflow. yes size-check is important and should block a merge, but not tests from running 😅

franky47:
Yeah I agree, feel free to bump it high enough (in nuqs' package.json) for it to pass and I'll see how to move it cleanly to another step (the docs build needs the actual size to display it on the landing page).

TkDodo:
I’ve increased the relevant size-limit for now a bit:

db3c11e2

packages/nuqs/src/useQueryStates.ts
```diff
@@ -383,20 +384,27 @@ function parseMap<KeyMap extends UseQueryStatesKeysMap>(
...
     const queuedQuery = queuedQueries[urlKey]
     const query =
       queuedQuery === undefined
-        ? (searchParams?.get(urlKey) ?? null)
+        ? parser.type === 'multi'
+          ? (searchParams?.getAll(urlKey) ?? [])
+          : (searchParams?.get(urlKey) ?? null)
         : queuedQuery
+    // todo this === comparison likely won't work with arrays
     if (cachedQuery && cachedState && (cachedQuery[urlKey] ?? null) === query) {
```
TkDodo: `cachedQueries` now contain `Iterable<string>`, so we can’t really compare them with triple equals.

I’m wondering why the `eq` method of the parser isn’t used for comparison here?

But, we could also extract this “shallow equals” array comparison to a shared util:

https://github.com/47ng/nuqs/blob/e8af13624d50e7461004c414d495bbaeb41c1f51/packages/nuqs/src/parsers.ts#L484-L492

TkDodo:
As I was trying to add some end-to-end tests, I’m stumbling over some issues I’m not quite sure how to fix / debug:

1. when multi-parsers are used, it would be great if `useQueryState` wouldn’t yield `null`, but always empty array instead. With `searchParams.getAll()`, it’s not possible to get null back.

I can see that `parseAsArray` also returns `Array<T> | null` so maybe that’s fine for multi parsers too, but I think when working with arrays, null is a bit annoying.

2. I’m not quite sure as to why yet, but when calling `setState([1, 2])` the url updates correctly but the state in the component reverts back to what it was from before the state update. It’s like it’s lagging behind or immediately gets re-set. Just leaving this here in case this sounds familiar to you :) I can commit the failing e2e test if you like.

packages/nuqs/src/useQueryStates.ts
franky47: > I’m wondering why the `eq` method of the parser isn’t used for comparison here?

We're comparing serialised versions here (the `eq` function compares state types), so yeah we'd likely have to branch for string comparison with `===` vs array comparison for iterables.

franky47:
>  I think when working with arrays, null is a bit annoying.

I agree, in a lot of cases `.withDefault([])` makes it easier. There have been cases where having a non-empty array as a default made sense for the domain, but then the URL would contain `?key=` for the empty array case, which is ugly and counter-intuitive (same as `parseAsBoolean.withDefault(true)`)

packages/e2e/shared/specs/native-array.tsx
```diff
@@ -0,0 +1,24 @@
+'use client'
+
+import { parseAsInteger, parseAsNativeArrayOf, useQueryState } from 'nuqs'
+import { Display } from '../components/display'
+
+export const parser = parseAsNativeArrayOf(parseAsInteger)
```
franky47, line 6: **suggestion**: Could we add `.withDefault([])` here to simplify the null-handling logic below?

packages/nuqs/src/lib/search-params.ts
```diff
@@ -0,0 +1,19 @@
+export function isEmpty(query: string | Iterable<string> | null): boolean {
+  return query === null || (Array.isArray(query) && query.length === 0)
```
franky47, line 4: **note**: Shouldn't this return true for the empty string case?

packages/nuqs/src/loader.ts
```diff
@@ -96,14 +97,18 @@ export function createLoader<Parsers extends ParserMap>(
...
+          ? searchParams.getAll(urlKey)
+          : searchParams.get(urlKey)
+      if (isEmpty(query)) {
         result[key] = parser.defaultValue ?? null
         continue
       }
       let parsedValue
       try {
-        parsedValue = parser.parse(query)
+        // we have properly narrowed `query` here, but TS doesn't keep track of that
```
franky47, line 110: **question**: Is it the `null` part that's causing issues? Could it be a matter of adding a type guard on `isEmpty` (like `: query is string | Iterable<string>`)

packages/e2e/shared/specs/native-array.tsx
TkDodo: good idea, but don’t you think it would make sense to build that into `parseAsNativeArrayOf` ?

> but then the URL would contain ?key= for the empty array case

interestingly, this doesn’t yet work in `parseAsNativeArrayOf`. If I set a default to `[42]` and then set the state explicitly to `[]`, the url is empty for me. This feels like a bug

packages/nuqs/src/lib/search-params.ts
TkDodo: the original logic wasn’t doing that - it was only checking for `===null`, in both places where I now use `isEmpty`:

https://github.com/47ng/nuqs/blob/270e60cdf243bdf12e60c16c3d2f8e9867c93b1a/packages/nuqs/src/loader.ts#L99-L100

https://github.com/47ng/nuqs/blob/270e60cdf243bdf12e60c16c3d2f8e9867c93b1a/packages/nuqs/src/useQueryStates.ts#L399

so the logic is: when the searchParam was retrieved with `.get(key)`, we want to compare with `null`, and when we do `.getAll(key)`, we need to compare with empty array.

packages/nuqs/src/loader.ts
TkDodo: no, null isn’t the problem, though the type-guard is a good idea anyhow. The problem is that when doing either `searchParams.getAll` or `searchParams.get`, the type of `query` is a union, ` string | Array<string>`.

the SingleParser needs a string, and the MultiParser needs an Array, however, we cannot prove that we only call a single parser wen we have done `.get` and a multi parser when we did `.getAll`.

we would likely need to completely switch the two conditions into two separate branches, I think this would work:

```
if (parser.type === 'multi') {
  const query = searchParams.getAll(urlKey)
  // ...
  parsedValue = parser.parse(query)
} else {
  const query = searchParams.get(urlKey)
  // ...
  parsedValue = parser.parse(query)
}
```

that would work but there is in-between code and the whole try/catch things that we would need to either duplicate or extract out and I didn’t think that would be worth it, but let me know if you prefer this direction

packages/e2e/shared/specs/native-array.tsx
franky47: > This feels like a bug

Indeed, the URL should only empty when setting the state to the default value (or `null` to clear it).

packages/nuqs/src/lib/search-params.ts
franky47: I see, maybe we could name this function `isAbsentFromUrl` or something similar, to remove ambiguity.

packages/nuqs/src/loader.ts
franky47: I see, whatever keeps the bundle size smaller (and works, obviously) is likely to be preferred. Having a comment for this kind of thing is fine.

packages/nuqs/src/lib/search-params.ts
TkDodo: see: ff107378

packages/nuqs/src/loader.ts
TkDodo: added the type-guard anyways: c8da7d1a

packages/e2e/shared/specs/native-array.tsx
TkDodo: added `withDefault`: 864e0cd8

and will look into the bug 🐛

packages/nuqs/src/useQueryStates.ts
TkDodo: I looked into this a bit more, but I’m a bit confused which types get stored where:

```
const query =
  queuedQuery === undefined
    ? parser.type === 'multi'
      ? (searchParams?.getAll(urlKey) ?? [])
      : (searchParams?.get(urlKey) ?? null)
   : queuedQuery
```

now when we read the Query from the `searchParams`, it can only be `string | null | Array<string>` - I’ve extracted that to its own type:

```
export type SearchParamQuery = string | Array<string>
```

but when I look at `queuedQuery`, it seems that we store whatever `parser.serialize` has returned, which used to only be strings, but now it can be `Iterable<string>`, so I’ve also made type for that:

```
export type SerializedQuery = string | Iterable<string>
```

now when trying to compare the two:

```
compareQuery(cachedQuery[urlKey] ?? null, query)
```

`cachedQuery` contains a `SerializedQuery`, and `query` can contain that too when we get it from `queuedQuery`, but if we read it directly from the `searchParams`, it will be a `SearchParamQuery`.

I don’t see how we could effectively compare the result of `searchParams.getAll()` with the result of a serialized Query, so maybe we need to run that trough `parser.serialize` before?

franky47: I see, if `.getAll()` returns a `Array<string>`, that is already an `Iterable<string>`, so maybe we could have a comparison function dedicated to serialised representations (either read from the URL or serialised before an update)

> so maybe we need to run that trough `parser.serialize` before?

The input of `serialize` being a state ("hydrated"/deserialised) data type, that wouldn't work, what we get from `searchParams.get{,All}()` is already in a serialised form.

The issue might be from the `?? null` part in `cachedQuery[urlKey] ?? null`, as this makes sense for a single parser, but not much for a multi one (where we'd default to an empty array).

TkDodo: > so maybe we could have a comparison function dedicated to serialised representations (either read from the URL or serialised before an update)

yeah can try that.

> The issue might be from the ?? null

does it even make sense to call the comparison function if `cachedQuery[urlKey]` doesn’t exist? this would never yield a cache hit imo...

TkDodo: untested, but here is an attempt at that:

e0021b19

franky47: One thing to look out for with cached queries is that `null` and `undefined` have different meanings: `undefined` means there is no cache in memory, but `null` here means we requested to remove that key from the URL when setting the state.

If you have suggestions for a better internal way of encoding these intents, I'm very much open to suggestions (maybe symbols?)

packages/e2e/shared/specs/native-array.tsx
TkDodo: I understand why this happens. When calling `write` here:

https://github.com/47ng/nuqs/blob/e8af13624d50e7461004c414d495bbaeb41c1f51/packages/nuqs/src/lib/queues/throttle.ts#L185-L189

the serialized value is an empty iterator, so it doesn’t append anything.

the fix is to explicitly append an empty string if we get an empty iterator. this seems weird, but when the default value is applied, we never call `write`, so the only way to get an empty array here is if we have a different default value (or no default value because the default is null) and then explicitly set to empty array.

That also means if you do:

```
parseAsNativeArrayOf(parseAsInteger)
```

without a default, and then call `setState([])`, you will see an empty param in the url. If you call `setState(null)`, it will be cleared. This is consistent imo, but it also means I really think we should have a built-in `.withDefault([])` on the native array multi parser to avoid that pitfall and help with null handling.

cc89e20b

packages/nuqs/src/useQueryStates.ts
TkDodo: yeah I kept that intent the same now (I think), and only fallback to `?? defaultValue` instead of `?? null`, where `defaultValue = parser.type === 'multi' ? [] : null`

TkDodo: but note that `??` will use the fallback for both null and undefined, so I’m not sure if that differentiation between null and undefined was there before in all places 🤔

TkDodo:
react-router v7 seems to fail because of this key isolation that does a `!==` check directly on `searchParams.get`:

https://github.com/47ng/nuqs/blob/489724856f95adc9b76c6eda3dd1645ae86a7eb0/packages/nuqs/src/adapters/lib/key-isolation.ts#L12

not sure if we can switch this conditionally to `getAll` because there’s no parser available anywhere 🤔

TkDodo:
we can switch to `compareQuery(oldValue.getAll(key), newValue.getAll(key))`, which would fix the issue, but it has a slightly different behaviour than before. if you compared:

```
?test=1&test=2
?test=1test=3
```

before, this would yield “no difference”, because it only looked at the first value it got. Now, we do get a difference.

on the other hand, if you’re using nuqs in your application, you probably don’t have the same value multiple times in the url ...

let’s see if that fixes it: c38d7804

TkDodo:
I also found the issue with TanStack Router: The default `decode` mechanism returns `{ test: 1}` if we have `?test=1` in the url, and `{ test: [1,2] }` when we have `?test=1&test=2`.

That’s usually fine, but the multi parser expects an array as input, so we can’t use the parser like this for `validateSearch`:

```
const validateSearch = createStandardSchemaV1(
  { test: parser },
  { partialOutput: true }
)
```

I guess we’d have the same problem the other way around: if we have multiple values in the url, `parser.serialize` of a SingleParser would be called with an array as input, which would die on certain parsers, e.g. `parseAsTimestamp`.

In fact, using this url:

```
?test="1970-01-01T00%3A00%3A01.970Z"&test="1970-01-01T00%3A00%3A01.970Z"
```

in TanStack Router with `parseAsTimestamp` as `validateSearch` leads to a runtime failure too:

```
"[nuqs] Failed to parse query `\"1970-01-01T00:00:01.970Z\",\"1970-01-01T00:00:01.970Z\"` for key `test` (got null)"
```

So, as it stands, with `createStandardSchema`, the assumption that the `serialize` method of a `parser` gets the type `T` passed is not correct.

I can try to make the `serialize` method of the multi-parser more defensive, but I think we need to actually build this into the serializer?

https://github.com/47ng/nuqs/blob/e8af13624d50e7461004c414d495bbaeb41c1f51/packages/nuqs/src/serializer.ts#L101

TkDodo:
this defensive check inside `parseAsNativeArrayOf` would fix it, but I think the underlying problem is that we pass something to `.serialize` that wasn’t produced by `.parse`

```diff
serialize: values => {
-  return values.flatMap(value => {
+  const safeValues = Array.isArray(values) ? values : [values]
+  return safeValues.flatMap(value => {
```

and according to the types, this shouldn’t be possible, so everyone who makes a custom parser will have the same problem.

I think we’d need to tell TanStack Router to use our `parser.parse` to read from the url, so maybe our tests are just missing that ?

franky47:
You might want to check out #1128 for the TanStack Router behaviours, that PR fixes a few things wrt/ serialization/parsing.

TkDodo:
sadly getting the same issue with those changes:

```
values.flatMap is not a function
```

as far as I can see, the TanStack adapter isn’t involved in this problem, this is mainly about how the values parsed by TanStack Router are passed to `validateSearch`, which we then pass to our `parser`.

TkDodo:
@franky47 I think the root cause is really this:

https://github.com/47ng/nuqs/blob/b9296827321fb10e2ede1b0acc22266d840a8b8f/packages/nuqs/src/standard-schema.ts#L43-L45

`input` that we get in is of type `unknown`, and we just pass it to `serialize` - the `as any` is a hint that we’re doing something that might not work at runtime.

looking at [the docs of createSerializer](https://nuqs.dev/docs/utilities#serializer-helper), it’s meant to be typed that way, but it also means it might not be suited for parsing arbitrary things :/

packages/nuqs/src/lib/search-params.ts
```diff
@@ -0,0 +1,24 @@
...
+  searchParams: URLSearchParams
+): URLSearchParams {
+  if (typeof serialized === 'string') {
+    searchParams.set(key, serialized)
+  } else {
+    searchParams.delete(key)
+    for (const v of serialized) {
+      searchParams.append(key, v)
+    }
+    if (searchParams.size === 0) {
```
ValeraS: Is this check correct? `searchParams` may contain values for other keys.
```suggestion
    if (!searchParams.has(key)) {
```

TkDodo: you’re right, the size check is probably wrong. nice catch 👍

packages/nuqs/src/lib/compare.ts
```diff
@@ -0,0 +1,33 @@
...
+): boolean {
+  if (a === b) {
+    return true // Referentially stable
+  }
+  if (a === null || b === null) {
+    return false
+  }
+  // we expect either strings or iterables, not a mix of both
+  if (typeof a === 'string' || typeof b === 'string') {
+    return a === b
```
franky47: **note**: This `a === b` here will always be `false` as it's been covered above.

franky47:
One issue of the widening of the `ParserBuilder` type (now encompassing `MultiParserBuilder`) is that it breaks composition of parsers, requiring to explicitly type nested parsers as `SingleParserBuilder`:

CleanShot 2025-09-19 at 23 04 56@2x

```ts
import { createParser, type ParserBuilder, useQueryState } from 'nuqs'

function kvEntriesParser<Item>(itemParser: ParserBuilder<Item>) {
  return createParser<[string, Item]>({
    parse(query) {
      const [key, value = ''] = query.split(':')
      return [key, itemParser.parse(value)]
    },
    serialize([key, value]) {
      return `${key}:${itemParser.serialize(value)}`
    },
    eq(a, b) {
      return a[0] === b[0] && itemParser.eq(a[1], b[1])
    }
  })
}
```

TkDodo:
right, we could keep everything "Parser" related as it is, naming wise, and name the union something like "GenericParser" instead of having Parser become SingleParser. I like the Single/Multi naming more but the other version is less breaking.

franky47:
We could mark the ParserBuilder type as deprecated with a note that it will change in v3, but keep the second option for non-breaking in v2.

Though `@deprecated` for a name that will remain, but change semantics is not ideal. What do you think?

packages/nuqs/src/lib/compare.ts
TkDodo: good point, fixed that here + added tests: f5bf1b84

TkDodo:
I think deprecating `Parser` and `ParserBuilder` is fine, I’ll do that and alias it to `SingleParser`. Additionally, the union type is `GenericParser` and `GenericParserBuilder`, though we can think about a better name. Ideally, it would be just `Parser`, but as you said, that changes the meaning, so I think the best way is to make `Parser` and `ParserBuilder` really go away in the next release.

see: 6b14b75d

packages/docs/content/docs/parsers/built-in.mdx
```diff
@@ -322,6 +323,38 @@ parseAsJson(userSchema.validateSync)
   return `null{:ts}` for invalid data. Only **synchronous** validation is supported.
 </Callout>
 
+## Native Arrays
+
+<FeatureSupportMatrix introducedInVersion='2.6.0' />
```
franky47: **suggestion**: 2.6.0 is already out, so this will land in 2.7.0.
```suggestion
<FeatureSupportMatrix introducedInVersion='2.7.0' />
```

packages/e2e/tanstack-router/src/routes/native-array.tsx
```diff
@@ -0,0 +1,13 @@
+import { createFileRoute } from '@tanstack/react-router'
+import { NativeArray, parser } from 'e2e-shared/specs/native-array'
+import { createStandardSchemaV1 } from 'nuqs'
+
+const validateSearch = createStandardSchemaV1(
```
franky47, line 5: One thing I noted in the TSR example:

If you play with the demo (add 3 items using the button), you get to a url of
```
http://localhost:3004/native-array/?test=1&test=2&test=3
```
which is fine. If then you reload the page, you end up with:
```
http://localhost:3004/native-array?test=%5B1%2C2%2C3%5D
```

Now for the even weirder part: if I **type manually** the first URL directly, the multi array remains in the URL, and survives reloads. But if I **copy/paste** it (exact same content), it gets immediately rewritten to the second form (encoded).

Same behaviour in Chrome, Firefox & Safari.

TkDodo: this is really weird. I was running nuqs in debug mode and all the logs indicate that we have an array with 3 items and URLSearchParams with a size of 3. Also, things just continue to work...

I guess this must have something to do with how TanStack Router serializes things into the url? It seems to take the 3 params and makes 1 out of them.

If I navigate to the same page and do:

```
http://localhost:3004/native-array/?foo=1&foo=2&foo=3
```

I get the same encoding situation. So it doesn’t seem to be nuqs related.

I also tried:

- http://localhost:3004/pretty-urls
- click the test button

the url becomes:

http://localhost:3004/pretty-urls/?test=-._~!$()*,;=:@/?[]{}\|^

now reload the page and the url becomes:

http://localhost:3004/pretty-urls?test=-._%7E%21%24%28%29*%2C%3B%3D%3A%40%2F%3F%5B%5D%7B%7D%5C%7C%5E

🤷

franky47: The trailing slash seems to matter:
- `http://localhost:3004/native-array?foo=1&foo=2&foo=3` → Doesn't encode, persists on reload
- `http://localhost:3004/native-array/?foo=1&foo=2&foo=3` → The trailing slash is removed, and search params are encoded

~I can reproduce this behaviour in a stock TSR app~, so unrelated to nuqs indeed.

packages/nuqs/src/testing.ts
franky47, line 1: @TkDodo I added utilities to test bijectivity of custom multi-parsers. The overloads should take care of retro-compatibility, but if you see some issue, let me know.

PR #1136, opened 2025-09-14
chore: split size-limit script from build process

### LINKED ISSUE #1135: Split size-limit check into its own CI step + size.json generation
Currently, the build step of the `nuqs` package includes a post-build script that runs size-limit.

This is cumbersome, as a failure here (going over the threshold) will prevent subsequent steps from running (like running tests).

There are two cases where we need to run `size-limit`:
1. To generate the `size.json` file that the docs need for displaying the bundle size on the landing page
2. To validate against a given threshold

Having the size.json generation as the package is built allowed the docs build process to only require building the nuqs package as a dependency.

We don't technically need to enforce the size while building the docs, and we don't need to emit the size.json when checking the size, so these two operations should be split into two scripts.

Then:
1. The docs build step could require a dependency on nuqs' size.json generation script (in turbo.json)
2. We could move the size checking (with failure when going over the threshold) to a CI step, that lets unit tests run (possibly also e2e tests, but blocking a deployment if going over the line)

## Related
- #1071 
- #1095 (cc @miinhho)

miinhho:
## Summary
Split size-limit operations into two separate scripts:

1. `build:size-json` - Only generates `size.json` for docs
2. `test:size` - Only validates size thresholds

Update `turbo.json` to:
- Remove `size.json` from `nuqs#build` outputs
- Add `nuqs#build:size-json` task
- Make `docs#build` depend on `nuqs#build:size-json`

#### `package.json`
- Removed postbuild script that ran `size-limit --json > size.json`
- Added `build:size-json` script: `size-limit --json > size.json`
- Kept `test:size` script: `size-limit` (for validation)

#### `turbo.json`
- Removed `size.json` from `nuqs#build` outputs
- Added `nuqs#build:size-json` task with `dependsOn: ["build"]`
- Updated `docs#build` dependsOn to include `nuqs#build:size-json`

## Testing
- ✅ `pnpm turbo run docs#build` succeeds with proper task order:
  - `nuqs:build` → generates dist/
  - `nuqs:build:size-json` → generates size.json
  - `docs:build` → uses both dist/ and size.json
- ✅ `pnpm run build:size-json` generates size.json independently
- ✅ `pnpm run test:size` validates size thresholds independently
- ✅ Bundle size display in docs landing page works correctly

Closes #1135.

### DIFF
--- a/packages/nuqs/package.json
+++ b/packages/nuqs/package.json
@@ -131,7 +131,7 @@
     "dev": "tsdown --watch",
     "prebuild": "rm -rf dist",
     "build": "tsdown",
-    "postbuild": "size-limit --json > size.json",
+    "build:size-json": "size-limit --json > size.json",
     "test": "pnpm run --stream '/^test:/'",
     "test:unit": "vitest run --typecheck",
     "test:size": "size-limit",
--- a/turbo.json
+++ b/turbo.json
@@ -10,7 +10,11 @@
       "dependsOn": ["^build"]
     },
     "nuqs#build": {
-      "outputs": ["dist/**", "size.json", ".tsup/**", ".tsbuildinfo"]
+      "outputs": ["dist/**", ".tsup/**", ".tsbuildinfo"]
+    },
+    "nuqs#build:size-json": {
+      "outputs": ["size.json"],
+      "dependsOn": ["build"]
     },
     "e2e-next#build": {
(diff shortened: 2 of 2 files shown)


franky47 (approved):
LGTM, thanks!

cc @TkDodo for #1134.

PR #1143, opened 2025-09-22
fix: optimistic state flicker

### LINKED ISSUE #1099: State temporarily resets to previous value on subsequent rerenders

franky47:
Adding tests to try and isolate the issue.

Fixes #1099.

### DIFF
--- a/packages/e2e/next/cypress/e2e/shared/repro-1099.cy.ts
+++ b/packages/e2e/next/cypress/e2e/shared/repro-1099.cy.ts
@@ -1,0 +1,25 @@
+import { testRepro1099 } from 'e2e-shared/specs/repro-1099.cy'
+
+testRepro1099({
+  path: '/app/repro-1099/useQueryState',
+  hook: 'useQueryState',
+  nextJsRouter: 'app'
+})
+
+testRepro1099({
+  path: '/app/repro-1099/useQueryStates',
+  hook: 'useQueryStates',
+  nextJsRouter: 'app'
+})
+
+testRepro1099({
+  path: '/pages/repro-1099/useQueryState',
+  hook: 'useQueryState',
+  nextJsRouter: 'pages'
+})
+
+testRepro1099({
+  path: '/pages/repro-1099/useQueryStates',
+  hook: 'useQueryStates',
+  nextJsRouter: 'pages'
+})
--- a/packages/e2e/next/next.config.mjs
+++ b/packages/e2e/next/next.config.mjs
@@ -5,6 +5,7 @@ const basePath =
 const config = {
   basePath,
   productionBrowserSourceMaps: true,
+  reactStrictMode: false,
   experimental: {
     clientRouterFilter: false,
     ...(process.env.REACT_COMPILER === 'true' ? { reactCompiler: true } : {}),
--- a/packages/e2e/next/src/app/app/repro-1099/useQueryState/page.tsx
+++ b/packages/e2e/next/src/app/app/repro-1099/useQueryState/page.tsx
@@ -1,0 +1,10 @@
+import { Repro1099UseQueryState } from 'e2e-shared/specs/repro-1099'
+import { Suspense } from 'react'
+
+export default function Page() {
+  return (
+    <Suspense>
+      <Repro1099UseQueryState />
+    </Suspense>
+  )
+}
--- a/packages/e2e/next/src/app/app/repro-1099/useQueryStates/page.tsx
+++ b/packages/e2e/next/src/app/app/repro-1099/useQueryStates/page.tsx
@@ -1,0 +1,10 @@
+import { Repro1099UseQueryStates } from 'e2e-shared/specs/repro-1099'
+import { Suspense } from 'react'
+
+export default function Page() {
+  return (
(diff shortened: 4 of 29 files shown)


packages/nuqs/src/adapters/next/impl.app.ts
```diff
@@ -67,16 +65,9 @@ export function useNuqsNextAppRouterAdapter(): AdapterInterface {
   const [optimisticSearchParams, setOptimisticSearchParams] =
     useOptimistic<URLSearchParams>(searchParams)
   const updateUrl: UpdateUrlFunction = useCallback((search, options) => {
-    const queuedThrottledKeys = globalThrottleQueue.reset()
```
georgwittberger, line 70: Removing this reset of the global throttle queue before the search params are updated is certainly a solution to the problem. However, as far as I understood the logic this was the only place in the App Router setup where the queue was reset once the params are flushed to the URL. This adapter provides `autoResetQueueOnUpdate: false`, so it will not reset the queue automatically. Without this reset here, will it still reset the queue in some other place I have missed?

franky47: Yes, the mutex system counts how many calls of the history API have been made, and resets the queue once all of Next.js' internal calls have been made. We had those two resets in a race conditions before.

franky47: You were right: the queue is indeed not reset 😓

I feel like this is the only way we can get both the 1099 and render-count tests to pass at the same time, reliably.

The queue is cleared on user navigation, so it doesn't grow infinitely and stays in sync with the URL, but it _feels_ wrong to leave it pending after the URL has updated. The issue being that:

- If we clear it too early, we get the 1099 fail (the optimistic state flickers out when a high-priority state update happens in parallel)
- If we clear it too late, we get an extra render

georgwittberger: Something that I was wondering about is: why do we always use a transition to update the URL? I assume this is necessary to make `useOptimistic` work as intended in case the option `shallow: false` is set.

But do we actually have to wrap all the update logic in `startTransition` if we have `shallow: true`?

My assumption is that the deprioritized state update in the transition is causing the flicker when there are other non-transition state updates involved in the component. Maybe you could try using `startTransition` only if `shallow: false` is set? Hopefully, this causes `useSearchParams` to return the updated URL search params directly on next render.

franky47: The issue was not much with the transition itself (we had that mechanism before the queue refactor in 2.5.0, for an issue that @aurorascharff found when pairing nuqs with startTransition & slow RSCs, see #718).

Since the refactor though, the optimistic state stored in uSES was either reset too early (causing the flicker) or too late (causing an extra re-render).

I found a better solution I think: reset the queue on the _next push_, just before we start loading it again. We now mark the queue for "ready to be reset" when resolving the returned Promise (URL has been updated, or at least requested to be). The next push occurs synchronously (same event loop tick) as the local, optimistic state update, and so the reset of the old pending query doesn't cause an extra render, and doesn't flicker as they both switch at the same time.

There may be cases where this falls apart, but as long as JavaScript is single-threaded, I think it might be the best solution until we refactor the core to a single store.

PR #1147, opened 2025-09-28
fix: don't bundle when running in esm.sh

### LINKED ISSUE #1145: esm.sh: Cannot read properties of null (reading 'useId')
## Context

```
-> Paste result from `cat package.json | grep -e nuqs` here

2.6.0
```

What framework are you using?

- ✅/❌ React SPA
 

## Description

vite config
```ts
import { defineConfig } from 'vite';
import react from '@vitejs/plugin-react';

// https://vite.dev/config/
export default defineConfig({
  plugins: [react()],
  resolve: {
    alias: {
          react: 'https://esm.sh/react',
          'react-dom': 'https://esm.sh/react-dom',
          nuqs: 'https://esm.sh/nuqs',
        },
  },
});

```
## Reproduction

https://stackblitz.com/edit/vitejs-vite-zauabfbt?file=src%2FApp.tsx

1. `npm run dev`
2. See error

 -->

franky47:
This causes the React peer dependency to be bundled alongside the nuqs code, causing duplicate React modules to be used, and with it the classic "cannot read {hookName} of null".

Fixes #1145.

### DIFF
--- a/packages/nuqs/package.json
+++ b/packages/nuqs/package.json
@@ -193,7 +193,6 @@
     "zod": "^4.1.5"
   },
   "esm.sh": {
-    "// note": "Don't bundle React in esm.sh - See https://github.com/47ng/nuqs/issues/1145",
     "bundle": false
   },
   "size-limit": [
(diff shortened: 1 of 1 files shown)


franky47:
Didn't seem to work.

Issue #760, opened 2024-11-13
Return value of `useQueryStates` does not immediately update after update in parsers


pdme:
## Context

What's your version of `nuqs`?
"nuqs": "^2.1.2"

What framework are you using?
NextJS

- ✅ Next.js (app router)

Which version of your framework are you using?
15.0.3

## Description
The return value of `useQueryStates` does not change if the parsers passed into it change.

## Reproduction

https://codesandbox.io/p/devbox/vkkl53

Example: Steps to reproduce the behavior:

1. Go to the codesandbox
2. Click on FLAG button
4. See that parsers change but return value not

 -->

franky47:
What are you trying to achieve that requires switching the parser type for a given key at runtime?

pdme:
I want to make the type of parser and its default value dependent on other parameters,. For example the user might trigger a setting that changes the nature of a certain query param, e.g. from single to multiple values, or having a different default value.

I can achieve this by sanitizing/parsing the values after the parsing by `useQueryStates`, but then I lose the default value behavior (back to `null`) and is frankly a bit cumbersome. Cleaner would be to be able to update the parsers at runtime...

pdme:
You can set the default value dynamically, but only once. You cannot change the default value at runtime, based on external factors (see https://codesandbox.io/p/devbox/vkkl53)...

In any case, thanks for your answer, and a great library in general.

franky47:
Ah, now this is indeed a bug with `useQueryStates`: changing the hook to useQueryState does follow a dynamic default value.

I'll take a look in #762, thanks!

pdme:
great, thanks so much. I'll take a look myself in the meantime as well, see if I can find the culprit...

franky47:
I think it has to do with the cache ref that's used for referential integrity, in useQueryState the default value is nullish-coalesced at the very end when returning the value, but in useQueryStates it's added to the cached state, and that isn't reactive to a change in the default value in the keyMap.

franky47:
I've got a fix that passes tests, though I'll run some more to see if it didn't break referential stability as a side-effect.

Can you try the following please?

```
pnpm add https://pkg.pr.new/nuqs@762
```

Please note that it only handles dynamic defaults, and not changing parser types at runtime.

pdme:
this seems to work! great job

PR #1148, opened 2025-09-29
feat: global default value store

TkDodo:
This PR takes the `defaultValue` implementation, which used to be local to a component instance, and hoists it to a “global store” - a simple `ref` that we pass down through context.

This store gets lazily initialized - when the defaultValue is first required. This ensures that multiple components that render in different places with different defaultValues still show the same data, as the second `defaultValue` simply gets discarded.

This PR is a prerequisite for allowing “default-value-as-a-function”, because functions can read from different places (like localstorage), which can change over time, so for consistency, we’d only want the first read.

This makes `defaultValue` in-line with how `useState` initializers work (though on a “global” level), and that’s also exactly how `intitialData` works in TanStack Query.


packages/nuqs/src/adapters/lib/context.ts
```diff
@@ -60,15 +64,34 @@ export type AdapterProvider = (
...
     )
+  }
+}
+
+export function useDefaultValueStore(): DefaultValueStore {
+  const context = useContext(defaultValueContext)
+  if (!context) {
+    throw new Error('[nuqs] No DefaultValueContext found')
+  }
+  return context
```
TkDodo, line 95: I thought about just initializing with empty object, but that might then be shared on the server between requests if it ever comes to that 🤔 .

Correct me if I’m wrong but a `NuqsAdapterProvider` is always needed to be added by consumers, right ?

packages/nuqs/src/useQueryState.test.tsx
```diff
@@ -129,6 +130,40 @@ describe('useQueryState: referential equality', () => {
...
+      return (
+        <div>
+          {id} value: {value}
+        </div>
+      )
+    }
+    const result = render(
+      <div>
+        <TestComponent id="first" defaultValue={5} />
+        <TestComponent id="second" defaultValue={23} />
```
TkDodo, line 156: note: this test fails on `main` because it renders first with 5 and second with 23. Now, it renders both with `5`, which I think (hope) is what we want.

packages/nuqs/src/useQueryStates.test.tsx
```diff
@@ -192,7 +192,7 @@ describe('useQueryStates: referential equality', () => {
     expect(result.current[0].str).toBe('foo')
     rerender({ defaultValue: 'b' })
     const [state] = result.current
-    expect(state.str).toBe('b')
+    expect(state.str).toBe('foo')
```
TkDodo, line 195: note: this change in test actually shows the (imo buggy) current behaviour: If a component re-renders with a different defaultValue, why would the state change? Not even uncontrolled inputs in react with a `defaultValue` work that way...

packages/nuqs/src/adapters/lib/context.ts
```diff
@@ -60,15 +64,34 @@ export type AdapterProvider = (
 export function createAdapterProvider(
   useAdapter: UseAdapterHook
 ): AdapterProvider {
-  return ({ children, defaultOptions, processUrlSearchParams, ...props }) =>
-    createElement(
+  return function NuqsAdapterProvider({
```
franky47, line 68: **note**: The NuqsAdapter could be nested in multiple places (with possibly different adapter-level defaults), I haven't tested this extensively, but it seemed to work.

In that case, would it make sense to pass an initial value to use for the defaultValueStore, so they can all use the same one? A bit like passing the same queryClient to multiple providers in TSQ.

packages/nuqs/src/useQueryStates.ts
```diff
@@ -100,17 +103,14 @@ export function useQueryStates<KeyMap extends UseQueryStatesKeysMap>(
...
-  const defaultValues = useMemo(
-    () =>
-      Object.fromEntries(
-        Object.keys(keyMap).map(key => [key, keyMap[key]!.defaultValue ?? null])
-      ) as Values<KeyMap>,
-    [
-      Object.values(keyMap)
-        .map(({ defaultValue }) => defaultValue)
-        .join(',')
-    ]
```
TkDodo: note: this `useMemo` dependency array being gone is a good thing because stringifying doesn’t work well with complex objects _and_ it wouldn’t work at all if we wanted to support default values as a function.

now, it’s no longer needed because the global store is “stable” and we actually only read the defaultValue imperatively, so it doesn’t need to be reactive.

packages/nuqs/src/useQueryStates.ts
```diff
@@ -191,7 +191,6 @@ export function useQueryStates<KeyMap extends UseQueryStatesKeysMap>(
   useEffect(() => {
     function updateInternalState(state: V) {
       debug('[nuq+ %s `%s`] updateInternalState %O', hookId, stateKeys, state)
-      stateRef.current = state
```
TkDodo: note: this update was redundant because `updateInternalState` is only called with `stateRef.current`, so at this point, that value is already correctly set.

packages/nuqs/src/useQueryStates.ts
```diff
@@ -200,7 +199,7 @@ export function useQueryStates<KeyMap extends UseQueryStatesKeysMap>(
           state,
           query
         }: CrossHookSyncPayload) => {
-          const { defaultValue } = keyMap[stateKey]!
+          const defaultValue = defaultValueStore[stateKey]
```
TkDodo, line 202: note: it might be a bit brittle here to only read from the `defaultValueStore`. We basically expect it to be initialized already at this point, which is true, but if we ever remove values from the store we would need to re-initialize here too.

packages/nuqs/src/adapters/lib/context.ts
TkDodo: I think we should keep this as an implementation detail as much as possible, so not expose anything. If you want, we can “carry over” defaultValues from existing contexts by just attempting a read instead of always initializing with an empty object.

This is how it could look like:

```diff
- const defaultValueStore = useRef({})
+ const defaultValueStore = useRef(useContext(defaultValueContext) ?? {})
```

it’s a simple change, and it would mean that even if you nest NuqsAdapters, you would always use a single defaultValue store.

packages/nuqs/src/adapters/lib/context.ts
franky47: > Correct me if I’m wrong but a NuqsAdapterProvider is always needed to be added by consumers, right ?

Correct.

packages/nuqs/src/useQueryState.test.tsx
franky47: I think this relates to the repro for issue #760 (dynamic defaults): a change in the default value should be reflected in the output state.

packages/nuqs/src/useQueryStates.test.tsx
franky47: The state internally can be nullable, and the default is nullish-coalesced on top of that, so that made the output respect dynamic defaults (issue #760).

packages/nuqs/src/adapters/lib/context.ts
franky47: Ah, inheriting from the parent, yes that's much better, I agree with keeping it internal.

packages/nuqs/src/useQueryState.test.tsx
TkDodo: I don’t think supporting dynamic default value is an intended feature - it feels more like a bug to me as it would result in different values being rendered on screen for the same key.

It also won’t work with the discussed `writeDefaults` feature.

packages/nuqs/src/adapters/lib/context.ts
TkDodo: done here: 50d35694

PR #1154, opened 2025-10-01
chore: add Knip for upkeeping

franky47:
- [x] Enable test coverage
- [x] Remove unused dependencies
- [x] Refactor some internals
- [x] Export `CreateStandardSchemaV1Options` type
- [x] Run knip in CI (needs all deps, maybe move it to its own job)
- [x] Enable knip for docs
- [x] Enable knip for e2e tests
- [x] Enable knip for examples


franky47:
@webpro I get a failure when running knip in CI (and can reproduce locally) with the following conditions:
- I install only the --workspace-root dependencies (where knip is defined)
- I run `pnpm run lint:knip --workspace packages/nuqs` to only check the nuqs package
- It fails when trying to load something unrelated in the e2e environments:

```shell
$ pnpm run lint:knip --workspace packages/nuqs

> nuqs-monorepo@0.0.0 lint:knip /Users/franky/dev/playground/47ng/nuqs
> knip --workspace packages/nuqs

ERROR: Error loading /Users/franky/dev/playground/47ng/nuqs/packages/e2e/react-router/v6/cypress.config.ts
Reason: Cannot find module 'e2e-shared/cypress.config'
Require stack:
- /Users/franky/dev/playground/47ng/nuqs/packages/e2e/react-router/v6/cypress.config.ts
 ELIFECYCLE  Command failed with exit code 2.
zsh: exit 2     pnpm run lint:knip --workspace packages/nuqs
```

Running `pnpm run lint:knip --workspace packages/nuqs` works fine if I have all the monorepo dependencies installed (although it takes a bit longer than when I had knip only set up as a dependency of the nuqs package).

Is knip trying to check somewhere it shouldn't?

webpro:
Thanks for giving Knip a shot! There are two things relevant here:

- When [linting a single workspace](https://knip.dev/features/monorepos-and-workspaces#lint-a-single-workspace), Knip includes ancestor and **dependent** workspaces. For instance, to figure out what exports of the linted workspace are unused. (It might be interesting to look into optimizations here to see exactly what operations could be skipped, but at the moment Knip calculates the relevant workspaces and does its thing on all of 'em and eventually reports only the issues for the provided workspace.)
- Knip is a _mostly_ static analysis tool, but it loads/imports configuration files like `cypress.config.ts`. That's because, in most cases, we want the _resolved_ exported configuration object to find entry points and dependencies. Some plugins have an (additional) AST resolver, but it's not as common.

The result is that dependencies of at least the ancestors and dependents should be installed upfront.

Does that help?

franky47:
That makes sense, thanks for the explanation! I'll run knip on a fully-installed dependencies setup then.

franky47:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

Issue #924, opened 2025-02-24
useQueryStates per field options not respected


pedro757:
## Context

What's your version of `nuqs`?

```
"nuqs": "^2.4.0",
```

What framework are you using?

- ✅/ Next.js (app router)

Which version of your framework are you using?

```
Operating System:
  Platform: linux
  Arch: x64
  Version: #1 SMP PREEMPT_DYNAMIC Thu Jan 23 21:29:10 UTC 2025
  Available memory (MB): 31502
  Available CPU cores: 4
Binaries:
  Node: 22.11.0
  npm: 10.9.0
  Yarn: 1.22.22
  pnpm: 9.14.2
Relevant Packages:
  next: 15.1.7 // Latest available version is detected (15.1.7).
  eslint-config-next: N/A
  react: 19.0.0
  react-dom: 19.0.0
  typescript: 5.7.3
Next.js Config:
  output: N/A
```

## Description

When using **useQueryStates** I would expect to have individual options, but setting `shallow: false` to one field will notify the server anyway when updating another field

## Reproduction

```tsx
"use client";

import React from "react";
import { parseAsString, useQueryStates } from "nuqs";

import { Button } from "@/components/ui/button";
import { Input } from "@/components/ui/input";

export default function VendedoresForm() {
  const [jsonData, setJsonData] = useQueryStates(
    {
      firstName: parseAsString.withDefault(""),
      lastName: parseAsString.withDefault("").withOptions({
        shallow: false,
      })
    },
    {
      shallow: true,
      history: "push",
      scroll: false,
    }
  );


  return (
    <form>
      <input
        onChange={e => setJsonData({ ...jsonData, firstName: e.target.value })}
        defaultValue={jsonData.firstName}
        placeholder="Ingresa el nombre"
      />
      <input
        onChange={e => setJsonData({ ...jsonData, lastName: e.target.value })}
        placeholder="Ingresa el apellido"
      />
    </form>
  );
}
```

Example: Steps to reproduce the behavior:

1. Update firstName field
2. See it notifies the server, even when that field does not have `{ shallow: false }` set
3. Remove `{ shallow: false }` from **lastName**
4. Now updating firstName will not notify the server as expected

franky47:
I think the issue comes from your setJsonData call.

[...]

PR #1160, opened 2025-10-04
perf: bail out of state updates for equal values

TkDodo:
When `useQueryStates` receives an event that should update its internal state. we should bail out of that update if we already store the same value.

This is what React’s `useState` does per default (by comparing with `Object.is`), however, it doesn’t work to delegate this to React because the internal state of `useQueryStates` is an object. When an event comes in, we only update a single value of that object:

```
stateRef.current = {
  ...stateRef.current,
  [stateKey as keyof KeyMap]: state ?? defaultValue ?? null
}
```

so we can optimize ourselves by comparing the current state for that key with the new state that comes in from the emit and skip calling `setInternalState` for those cases.

### DIFF
--- a/packages/nuqs/src/useQueryStates.ts
+++ b/packages/nuqs/src/useQueryStates.ts
@@ -189,11 +189,6 @@ export function useQueryStates<KeyMap extends UseQueryStatesKeysMap>(
 
   // Sync all hooks together & with external URL changes
   useEffect(() => {
-    function updateInternalState(state: V) {
-      debug('[nuq+ %s `%s`] updateInternalState %O', hookId, stateKeys, state)
-      stateRef.current = state
-      setInternalState(state)
-    }
     const handlers = Object.keys(keyMap).reduce(
       (handlers, stateKey) => {
         handlers[stateKey as keyof KeyMap] = ({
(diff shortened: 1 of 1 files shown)


franky47:
Could this be a solution to https://github.com/47ng/nuqs/issues/924#issuecomment-3289460479 ?

TkDodo:
> Could this be a solution to [#924 (comment)](https://github.com/47ng/nuqs/issues/924#issuecomment-3289460479) ?

Hm, I don't think so. The bailout happens later, only for setting component state. And with clearOnDefault: false, we might actually need to write to the url even though we set to the same value in state if it's equal to the default.

TkDodo:
@franky47 do you know what I can do against the react-router v7 test failure? I think this is a shared test, but it seems that we have one fewer re-renders now (which is a good thing)

franky47:
I'll run it again, the render test can be flaky sometimes, I don't like it but it has been useful in detecting major re-render changes in the past.

TkDodo:
I also need to add another test that calls `setState(v => v)` to confirm the behaviour compared to `main`

franky47:
The one change I see that could happen here is that a `setState(v => v, { shallow: false })` could have been used before to refresh server-side data, but there might be better (more explicit) ways to do this in userland.

TkDodo:
I actually messed up the last refactoring and the test showed it 🎉 

Should be good now: f17fe594

franky47:
I think there was a place in the docs or the playground (need to dig) where I explored a pattern to collect state changes locally, without updating the URL (to avoid polluting the global history), and submit all in one go by using this `setState(v => v)` trick.

To be fair, the `throttleMs: Infinity` trick to disengage the URL update is likely to cause some more issues with `writeDefaults`, and the pattern described above is a form, really. #UseThePlatform

TkDodo:
hmm I 'm not seeing that anywhere 🤔

franky47:
Found it, it's the deferred test in the e2e-next environment (which could be a shared one), and it still passes apparently:
https://github.com/47ng/nuqs/blob/next/packages/e2e/next/src/app/app/deferred/page.tsx

franky47 (approved):

PR #1163, opened 2025-10-06
fix: make sure the state-change effect runs for multi parsers

### LINKED ISSUE #1162: When navigating using forward/back 'parseAsNativeArrayOf' does not trigger state-change
## Context

```
2.7.0
```

What framework are you using?

- ✅ React Router

```
        "react-router-dom": "6.11.2"
```

## Description

Just updated to 2.7.0 to make use of the 'parseAsNativeArrayOf' parser since it better aligns with my backend.

My toggle-button-group allows for multiple selections, and when using the buttons to click, the url updates correctly, same with deselecting. 

My issue is when i use the 'history: "push"' option to be able to use the back button for previous states, changes are not emitted. I have checked and if i use 'parseAsArrayOf' the state updates correctly using forward/back navigation.

TkDodo:
fixes #1162 

the dependency array of the state-change effect included stringified versions of searchParams, but it was always using `.get`, which was only seeing the first search param.

With multi-parsers, we need to look at _all_ searchParams. It should be safe here to always call `.getAll()`, because even for one key, this would be stable.


packages/nuqs/src/useQueryStates.ts
```diff
@@ -153,10 +153,15 @@ export function useQueryStates<KeyMap extends UseQueryStatesKeysMap>(
...
-      ])
+      Object.entries(resolvedUrlKeys).map(([key, urlKey]) => {
+        const parser = keyMap[key]
+        return [
+          urlKey,
+          parser?.type === 'multi'
+            ? initialSearchParams?.getAll(urlKey)
+            : (initialSearchParams?.get(urlKey) ?? null)
+        ]
+      })
```
TkDodo, line 164: here, I’m trying to address a caching issue where we were putting the wrong values into the `cachedQuery` ( = `queryRef.current`) cache for multi parsers, because we were only reading the first search param with `.get`

Now, we are reading with `.getAll` for multi-parsers

@franky47 I’m not really sure how / where to add a test for this. I did add an e2e test for issue #1162

packages/e2e/shared/specs/native-array.cy.ts
```diff
@@ -17,4 +17,17 @@ export const testNativeArray = createTest('parseAsNativeArray', ({ path }) => {
...
+    cy.contains('#hydration-marker', 'hydrated').should('be.hidden')
+    cy.get('#client-name').should('be.empty')
+    cy.get('#add-button').click()
+    cy.get('#add-button').click()
+    cy.get('#add-button').click()
+    cy.location('search').should('eq', '?test=1&test=2&test=3')
+    cy.get('#client-name').should('have.text', '1 - 2 - 3')
+    cy.go('back')
+    cy.location('search').should('eq', '?test=1&test=2')
+    cy.get('#client-name').should('have.text', '1 - 2')
```
franky47, line 31: **suggestion**: There is technically no reason why this wouldn't be the case, but could we add a cy.go('forward') here and check we're back in the initial state?

packages/nuqs/src/useQueryStates.ts
franky47: Yeah the e2e test should be enough for this, thanks!

packages/e2e/shared/specs/native-array.cy.ts
TkDodo: 🫡: 7222c225

PR #1171, opened 2025-10-13
doc: support markdown response for AI

aryasaatvik:
## 🤖 LLM-Friendly Documentation Support

This PR implements AI agent-friendly documentation features, addressing maintainer feedback for better performance and cleaner URLs.

### ✨ Key Features

**Content Negotiation**
- Static rewrites based on Accept headers (no middleware overhead)
- Detects AI agents via `text/markdown` or `text/plain` without `text/html`
- Clean URL structure: `/llms/docs/` instead of `/docs/llms.mdx/`

**Multiple Endpoints**
- `/llms-full.txt` - Complete documentation corpus for LLMs
- `/llms/docs/[...]` - Individual page markdown content
- `/docs/[...]` - Regular HTML documentation

**Enhanced UX**
- LLM Copy Button and View Options for easy markdown access
- Page actions integrated into documentation layout

### 🔧 Technical Changes

- **Static Rewrites**: Accept header-based content negotiation in `next.config.mjs`
- **Route Handlers**: Created `/llms/docs/[[...slug]]` for markdown content
- **API Endpoints**: Added `/llms-full.txt` and `/llms.txt` for corpus access
- **Dependencies**: Updated Fumadocs (core: 15.8.5, mdx: 12.0.3)
- **Configuration**: Enabled `includeProcessedMarkdown` for LLM text extraction

### ✅ Addresses Maintainer Feedback

- **Performance**: Eliminated middleware runtime overhead
- **URL Structure**: Implemented clean `/llms/docs/` pattern  
- **TypeScript**: Fixed dependency version mismatches

### 🧪 Testing

```bash
# AI agent (gets markdown)
curl -H "Accept: text/markdown, text/plain" /docs/installation

# Browser (gets HTML)
curl -H "Accept: text/html" /docs/installation
```

### 📚 References

- [Fumadocs PR #2408](https://github.com/fuma-nama/fumadocs/pull/2408)
- [Bun JavaScript](https://x.com/bunjavascript/status/1971934734940098971)
- [Mintlify](https://x.com/mintlify/status/1972315377599447390)
- [SST OpenCode](https://github.com/sst/opencode/commit/c148f10bbd0c5a59160050218bcc850c9fb58056)

### DIFF
--- a/packages/docs/src/app/llms-full.txt/route.ts
+++ b/packages/docs/src/app/llms-full.txt/route.ts
@@ -4,8 +4,13 @@ import { getLLMText } from '@/src/lib/get-llm-text';
 // cached forever
 export const revalidate = false;
 
+const PAGE_EXCLUSIONS = [
+  "/docs/about",
+];
+
 export async function GET() {
-  const scan = source.getPages().map(getLLMText);
+  const pages = source.getPages().filter(page => !PAGE_EXCLUSIONS.includes(page.url));
+  const scan = pages.map(getLLMText);
   const scanned = await Promise.all(scan);
 
   return new Response(scanned.join('\n\n'));
--- a/packages/docs/src/lib/get-llm-text.ts
+++ b/packages/docs/src/lib/get-llm-text.ts
@@ -1,15 +1,7 @@
 import type { source } from "@/src/app/source";
 import type { InferPageType } from "fumadocs-core/source";;
 
-const PAGE_EXCLUSIONS = [
-  "/docs/about",
-];
-
 export async function getLLMText(page: InferPageType<typeof source>) {
-  if (PAGE_EXCLUSIONS.includes(page.url)) {
-    return null;
-  }
-
   const processed = await page.data.getText('processed');
 
   return `# ${page.data.title} (${page.url})


packages/docs/src/middleware.ts
```diff
@@ -0,0 +1,20 @@
+import { NextRequest, NextResponse } from "next/server";
+import { isMarkdownPreferred, rewritePath } from "fumadocs-core/negotiation";
+
+const { rewrite: rewriteLLM } = rewritePath("/docs/*path", "/docs/llms.mdx/*path");
+
+export function middleware(request: NextRequest) {
+	if (isMarkdownPreferred(request)) {
```
franky47: **question**: What does this function do?

aryasaatvik: it checks the Accept header for the presence of text/plain or text/markdown preference and the order and q value if both text/html and (text/plain or text/markdown) are present. 

using a static rewrite also works for claude code and opencode since they dont rely on q value and order

packages/docs/package.json
```diff
@@ -18,7 +18,7 @@
     "build": "next build",
     "start": "next start",
     "postinstall": "fumadocs-mdx",
-    "test": "tsc"
+    "test": "next typegen && tsc"
```
aryasaatvik, line 23: added next typegen in the docs#test command. [RouteContext](https://github.com/47ng/nuqs/pull/1171/files#diff-712f6310c7b58df52a44046c312d68de92030f7321f8429b43d15f40600ad3a9R9) needs to be generated using next build or next typegen before type checking

https://nextjs.org/docs/app/api-reference/config/typescript#route-aware-type-helpers

packages/docs/src/lib/cn.ts
```diff
@@ -0,0 +1 @@
+export { twMerge as cn } from 'tailwind-merge';
```
franky47: **note**: we have `import { cn } from '@/src/lib/utils'` throughout the codebase, can we use that one instead?

packages/docs/src/components/page-actions.tsx
```diff
@@ -0,0 +1,247 @@
...
+              strokeLinejoin="round"
+            />
+          </svg>
+        ),
+      },
+      {
+        title: 'Open in ChatGPT',
+        href: `https://chatgpt.com/?${new URLSearchParams({
+          hints: 'search',
+          q,
```
franky47: **suggestion**: use nuqs' serializer here (dogfood FTW 😁)

packages/docs/src/components/page-actions.tsx
```diff
@@ -0,0 +1,247 @@
...
+        href: githubUrl,
+        icon: (
+          <svg fill="currentColor" role="img" viewBox="0 0 24 24">
+            <title>GitHub</title>
+            <path d="M12 .297c-6.63 0-12 5.373-12 12 0 5.303 3.438 9.8 8.205 11.385.6.113.82-.258.82-.577 0-.285-.01-1.04-.015-2.04-3.338.724-4.042-1.61-4.042-1.61C4.422 18.07 3.633 17.7 3.633 17.7c-1.087-.744.084-.729.084-.729 1.205.084 1.838 1.236 1.838 1.236 1.07 1.835 2.809 1.305 3.495.998.108-.776.417-1.305.76-1.605-2.665-.3-5.466-1.332-5.466-5.93 0-1.31.465-2.38 1.235-3.22-.135-.303-.54-1.523.105-3.176 0 0 1.005-.322 3.3 1.23.96-.267 1.98-.399 3-.405 1.02.006 2.04.138 3 .405 2.28-1.552 3.285-1.23 3.285-1.23.645 1.653.24 2.873.12 3.176.765.84 1.23 1.91 1.23 3.22 0 4.61-2.805 5.625-5.475 5.92.42.36.81 1.096.81 2.22 0 1.606-.015 2.896-.015 3.286 0 .315.21.69.825.57C20.565 22.092 24 17.592 24 12.297c0-6.627-5.373-12-12-12" />
+          </svg>
+        ),
+      },
+      {
+        title: 'Open in Scira AI',
```
franky47, line 135: **note**: I've never heard of Scira, but I see Fumadocs also has it on their docs.

packages/docs/src/app/docs/[[...slug]]/page.tsx
```diff
@@ -28,7 +29,16 @@ export default async function Page(props: PageProps) {
...
       <DocsTitle>{page.data.title}</DocsTitle>
-      <DocsDescription>{page.data.description}</DocsDescription>
+      <DocsDescription>{page.data.description}
+        <div className="flex flex-row gap-2 items-center border-b pt-2 pb-6">
+          <LLMCopyButton markdownUrl={`${page.url}.mdx`} />
+          <ViewOptions
+            markdownUrl={`${page.url}.mdx`}
+            githubUrl={`https://github.com/47ng/nuqs/blob/dev/packages/docs/content/docs/${page.path}`}
+          />
+        </div>
```
Comment (franky47) [inline packages/docs/src/app/docs/[[...slug]]/page.tsx]: **suggestion**: These are client components at the top of the page, they probably should have a Suspense boundary with a skeleton fallback to avoid layout shifts on slow connections.

franky47:
Looking at the [output of `/llms.txt`](https://nuqs-22yf2iyvn-47ng.vercel.app/llms.txt), I see a lot of "noise", like the about page coming first which doesn't bring much value to the context.

Also, the output of llms.txt and llms-full.txt appear to be the same.

I wonder if we could fence off human-only content with HTML comments, like:
```md
content for both humans & llms
<!-- llms:off -->
human only content
<!-- llms:on -->

we could also have LLM-only contents like so:

<!-- llms:only
this is not rendered for humans, but emitted in llms.txt and other markdown endpoints.
-->
```

aryasaatvik:
> Looking at the [output of `/llms.txt`](https://nuqs-22yf2iyvn-47ng.vercel.app/llms.txt), I see a lot of "noise", like the about page coming first which doesn't bring much value to the context.
> 
> Also, the output of llms.txt and llms-full.txt appear to be the same.
> 
> I wonder if we could fence off human-only content with HTML comments, like:
> 
> ```
> content for both humans & llms
> 
> human only content
> 
> 
> we could also have LLM-only contents like so:
> 
> 
> ```

updated llms.txt to only include a table of contents

also added support for page exclusions for llms-full

I think fencing content for llms/humans in .mdx would require creating a fumadocs loader plugin to transform the comments at the mdx compilation step and filtering content. i don't want to do this as a part of this PR

franky47:
Closing in favour of #1271.

PR #1175, opened 2025-10-17
feat(docs): add TanStack Table parsers

Popescu-PfeifferMarc:
Add a new section in the docs covering how to use custom parsers with TanStack Table for filtering and sorting. The current implementation is somewhat limited as it only supports sorting and filtering one column at a time.

Please note that this PR was implemented by using Claude Code to adjust code copied over from another codebase of mine to the format which the docs expect and I have not reviewed each change in detail. Feel free to close if this is not acceptable. I just thought to leave this here if anyone else searches for it or to at least provide a starting point.

### DIFF
--- a/packages/docs/content/docs/parsers/community/tanstack-table.generator.tsx
+++ b/packages/docs/content/docs/parsers/community/tanstack-table.generator.tsx
@@ -23,7 +23,9 @@ import {
   parseAsIndex,
   parseAsInteger,
   parseAsString,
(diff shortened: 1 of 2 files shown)


franky47:
Hey, thanks, I don't know if this is going to be merged because there already are a few PRs about this, and I think we might end up shipping some variant of TST parsers as part of the upcoming shadcn CLI integration.

Issue #1118, opened 2025-09-05
Custom registry installation support using shadcn CLI instead of copy-pasting for community parsers


franky47:
### Discussed in https://github.com/47ng/nuqs/discussions/916

<div type='discussions-op-text'>

<sup>Originally posted by **junwen-k** February 13, 2025</sup>
We could potentially leverage shadcn's new custom [registry](https://ui.shadcn.com/docs/registry) support for a seamless community parsers installation process. I ran a quick test, and it seems to work:

`registry.json`
```json
{
  "$schema": "https://ui.shadcn.com/schema/registry.json",
  "name": "nuqs",
  "homepage": "https://nuqs.47ng.com",
  "items": [
    {
      "name": "use-pagination-search-params",
      "type": "registry:hook",
      "title": "usePaginationSearchParams",
      "description": "A hook to use pagination search params.",
      "dependencies": ["nuqs"],
      "files": [
        {
          "path": "registry/hooks/use-pagination-search-params.ts",
          "type": "registry:hook"
        }
      ]
    }
  ]
}
```

`registry/hooks/use-pagination-search-params.ts`
```ts
import { createParser, parseAsInteger, useQueryStates } from "nuqs";

// Page index is zero-indexed internally but one-indexed in the URL
// to align with UI expectations.
const pageIndexParser = createParser({
  parse: (query) => {
    const page = parseAsInteger.parse(query);
    return page === null ? null : page - 1;
  },
  serialize: (value) => parseAsInteger.serialize(value + 1),
});

const paginationParsers = {
  pageIndex: pageIndexParser.withDefault(0),
  pageSize: parseAsInteger.withDefault(10),
};

const paginationUrlKeys = {
  pageIndex: "page",
  pageSize: "perPage",
};

export function usePaginationSearchParams() {
  return useQueryStates(paginationParsers, { urlKeys: paginationUrlKeys });
}
```

### Installation
Following the [shadcn setup guide](https://ui.shadcn.com/docs/registry/getting-started), after running:

```zsh
shadcn build
```

We get the registry output. Now, assuming this is hosted at `https://nuqs.47ng.com/r/use-pagination-search-params.json`, we can seamlessly install the hook with:

```zsh
pnpm dlx shadcn@latest add https://nuqs.47ng.com/r/use-pagination-search-params.json
```

This might however require users to have the `components.json` in order for it to work though, but it's worth exploring!</div>

## Scoped & trusted registries

[...]

PR #1188, opened 2025-10-25
feat: Shadcn Registry Package

### LINKED ISSUE #1022: parseAsTuple

I-3B:
closes #1022 #1036 #1118 

## Checklist 
- [x] Registry package
- [x] `parseAsTuple` registry item
- [x] unit tests for `parseAsTuple`
- [x] add usage .md for `parseAsTuple`


franky47:
Thanks! I was working on the docs part (listing of installable items and copy/paste vs CLI methods) while on the plane, I'll open a PR later today.

I-3B:
hi @franky47 !
this is still WIP, but would like your opinion on exposing `lib/` like this
also I'm not entirely sure about the trusted registry part, do we need to submit anything to shadcn?

franky47:
Yeah I'll open a PR to the shadcn registry, as we have the [green light from the man himself](https://x.com/shadcn/status/1982025984401404063), but once #1189 is live.

I have opened #1189 as a draft too for the docs UI part, we could stack yours on top of that one to add parseAsTuple.

At a first glance, I'm not super keen on widening the exported API for internal purposes, I'll have to take a closer look once I'm home. But in #1189 I have all the "sources" for the registry as part of where it is deployed (in the docs package), and that causes problems in terms of dependencies, so a separate package might make more sense.

Like I wanted to add the equivalent of the next-typed-links snippet for React Router's href utility, but that would have implied adding the react-router dependency to the docs, along with tests harnesses etc.. Better do a separate package, have it build the registry JSON files and mark it as a dependency of the docs. We'll still need to copy the built output into the docs' `public/r` directory (having root level item URLs might clash and the trusted registry mapping makes the `/r/` subpath invisible anyway) before build starts (as the registry docs page read those files to populate its content).

I-3B:
> At a first glance, I'm not super keen on widening the exported API for internal purposes

agree, alternative is to add a `lib/utils.ts` (like shdacn's `cn` function), but didn't go that way cause it will mean there will be two copies of this util to maintain (which i think is okay, it's not a big function anyway)

packages/registry/lib/nuqs-safe-parse.ts
I-3B, line 1: I didn't add debug file as I thought it's an internal package detail that shouldn't be put in users' files, tradeoff of course is that debug flag doesn't work with community parsers

packages/registry/registry.json
```diff
@@ -0,0 +1,21 @@
...
+      "name": "parseAsTuple",
+      "type": "registry:file",
+      "title": "parseAsTuple",
+      "description": "Parse query string as a tuple of values.",
+      "dependencies": ["nuqs"],
+      "files": [
+        {
+          "path": "parsers/parseAsTuple.ts",
+          "type": "registry:file",
+          "target": "~/components/parsers/parseAsTuple.ts"
```
franky47: **note**: This would also need to install the safeParse utility as it's used in the parseAsTuple file.

Maybe it would make sense to export it from nuqs then, as it's useful for custom parsers anyway.

franky47 (changes requested):
I merged #1189, feel free to rebase on top of that.

It might make sense to expose the `safeParse` function as initially suggested (which would also make the `feat:` prefix a legitimate minor bump for the nuqs package).

packages/registry/items/parse-as-tuple.md
franky47, line 1: **suggestion**: usage files should remain next to their source files.

packages/docs/package.json
```diff
@@ -17,7 +17,7 @@
     "dev": "next dev --turbo",
     "build": "pnpm run build:registry && pnpm run build:next",
     "build:next": "next build",
-    "build:registry": "shadcn build",
+    "build:registry": "shadcn build --cwd ../registry --output ../docs/public/r",
```
franky47: **suggestion**: the new "assemble" script can deal with pulling items from the registry package, and so the shadcn build command can then use the default arguments.

packages/registry/lib/typed-links.ts
franky47, line 1: **note**: This one would be difficult to test without having a proper Next.js setup to typegen route definitions from, which we do have in the docs and could test against there.

packages/registry/registry.json
franky47, line 1: **note**: This registry file wouldn't be needed anymore: each item would need an individual `{name}.json` following the item schema (and an optional `{name}.md` for the usage docs).

packages/docs/package.json
franky47, line 1: **note**: we would also need to add the registry package as a dependency of the docs package, so we can go read its items from the symlinked `node_modules`.

franky47 (changes requested):
Thanks, there has been a few changes in how the registry is built in #1209:

- Each item is now described in an individual `packages/docs/src/registry/{item-name}.json`
- Next to that file can be an optional `{item-name}.md` for usage docs
- Those items _can_ specify a URL for file paths, in which case its contents will be fetched when building the registry (we use this for the Inertia adapter which lives in a [PHP example monorepo](https://github.com/47ng/nuqs-inertia-pingcrm)).

I think we could modify the "assemble" script that pulls all these items together into a generated root `registry.json` file to also go read from `packages/docs/node_modules/registry/items/**/*.json` (symlinked to `packages/registry/items/**/*.json`) or something similar, to enforce good boundaries between packages: registry for testing items, docs for deployment.

I thing there would still be cases where items could live:
- In the docs directly, if simple enough, or if an integration with Next.js is needed (eg: typed links)
- Outside the repo, if the test harness pulls too many dependencies that would slow down CI (eg: adapters for One.js or Expo Router, we might not want to add a full React Native stack to this monorepo just for one copy-pastable block of code).

But there is a lot of value in having items living in a testable registry package (or set of packages if we see in the future that some test harnesses conflict with one another).

I'm going to be preparing for a talk at React Advanced at the end of the week, so I might not be super active in PRs, but if you need guidance in how to adapt this PR to fit that model, feel free to ping me here. I've left some comments here and there too.

I-3B:
Thanks for the review and good luck with the talk!

> In the docs directly, if simple enough, or if an integration with Next.js is needed (eg: typed links)

even if the item is simple enough, we still would want to have it tested, yes? (e.g. parseAsTuple), so either we add tests right in docs package, or move all items, even simple ones, to registry package as they need to be tested.

My worry is that it will be hard to maintain 2 resources of registry items with no clear distinction on why a particular item is under docs or registry.

so need to decide if we want to discard the registry package altogether, and keep everything under docs (with their tests), or move all items to be under registry package, and update assemble script to pull from registry package.

or something else?

franky47:
I was thinking of the nextjs-typed-routes, which pulls type definitions from the `next` package, which are generated. This one could leverage the setup in the docs, or we could replicate a test setup in the registry package.

I-3B:
I see, okay then as a start I'll update the registry to work with the assemble script, might also move the current registry items that are under docs to be under registry (except nextjs-typed-routes).

packages/registry/items/parse-as-tuple.md
```diff
@@ -0,0 +1,12 @@
...
+
+```ts
+import { parseAsTuple } from '@/lib/parsers/parse-as-tuple'
+import { parseAsInteger } from 'nuqs'
+
+// Coordinates tuple (x, y)
+parseAsTuple([parseAsInteger, parseAsInteger])
+
+// Optionally, customise the separator
+parseAsTuple([parseAsInteger, parseAsInteger], ';')
```
franky47: **suggestion**: we could showcase a combination of different parsers in those docs, like `[string, 'asc' | 'desc']` for a sorting tuple or something similar.

packages/registry/items/parse-as-tuple.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "parse-as-tuple",
+  "type": "registry:item",
+  "title": "Tuples",
+  "description": "Parse query string as a tuple of values.",
+  "dependencies": ["nuqs"],
+  "files": [
+    {
+      "path": "node_modules/registry/items/parse-as-tuple.ts",
```
franky47, line 9: **question**: I'm pondering about casing in the naming of items, particularly parsers: `@nuqs/parse-as-tuple` vs `@nuqs/parseAsTuple`.

We had this discussion in https://github.com/47ng/nuqs/pull/1216#discussion_r2529064746, but it feels like if both of you instinctively went for `parse-as-whatever`, maybe it makes more sense, what do you think?

franky47 (commented):
Thanks, the registry dependency looks much better. However, it looks like the usage markdown file wasn't picked up:

Image

Edit: yeah the readUsage function still expects it to be located in the docs registry directory:
https://github.com/47ng/nuqs/blob/next/packages/docs/src/registry/read.ts#L35-L36

packages/registry/items/parse-as-tuple.json
I-3B: I decided on parse-as-tuple because 1. codebase is also kebab-case 2. shadcn also follow this convention for its components, so it more sense to me this way.

franky47: Sounds good.

PR #1197, opened 2025-11-02
fix: TypeError in serializer for array with default value

janglad:
Previously 

```ts
const serialize = createSerializer({
	arr: parseAsArrayOf(parseAsString).withDefault([])
 })
 ```

would throw when serialising with `{ arr : null } `. 

```
TypeError: Cannot read properties of null (reading 'length')
 ❯ eq src/parsers.ts:504:13
    502|         return true // Referentially stable
    503|       }
    504|       if (a.length !== b.length) {
       |             ^
    505|         return false
    506|       }
```

From what I understand this should be allowed, and is indicated as such in the types. This was caused by the equality fn being called with null. This PR fixes this and adds a test to cover this case.

### CODE REVIEW

**Comment by franky47** [comment]
I need to work on the CI part of things to follow NPM's new security directives (OIDC trusted publishing), as they will deprecate long-lived deployment tokens soon, then this should be good to go live.

Feel free to use the pkg.pr.new deployment if you want to use the hotfix in the mean time:
```shell
pnpm add https://pkg.pr.new/nuqs@bec3404
```

PR #1204, opened 2025-11-06
fix: make parseAsBoolean case insensitive

### LINKED ISSUE #1203: parseAsBoolean parses TRUE as false
## Context

```
2.7.2
```

What framework are you using?

- ✅ Next.js (app router)

```
15.4.3
```

## Description

My application interacts with existing systems that use 'TRUE' rather than 'true' for booleans in search params. I was a little surprised to find that nuqs parses 'TRUE' as false.

## Reproduction

```
parseAsBoolean.parse('true')  // returns true
parseAsBoolean.parse('TRUE')  // returns false 
```

## Proposed fix
See draft PR #1204

joshuaobrien:
Closes #1203.


franky47 (approved):
LGTM, thanks!

PR #1216, opened 2025-11-14
feat: add nuqs-presets pattern hooks to registry

iHiteshAgrawal:
Adds 7 production-ready pattern hooks for common URL state patterns:
- use-pagination: Complete pagination with page size control
- use-search: Debounced search with validation
- use-sorting: Multi-column sorting with smart toggling
- use-filters: Type-safe multi-filter management
- use-multi-select: Array-based multi-selection
- use-tabs: Type-safe tab navigation
- use-date-range: Date range selection with presets

All hooks maintained at: https://github.com/iHiteshAgrawal/nuqs-presets


packages/docs/src/app/playground/debug-control.tsx
```diff
@@ -7,6 +7,9 @@ export function DebugControl() {
     if (typeof localStorage === 'undefined') {
       return false
     }
+    if (typeof localStorage.getItem !== 'function') {
+      return false
+    }
```
franky47: **question**: Is this an issue in Safari? We had a [discussion on Bluesky](https://bsky.app/profile/francoisbest.com/post/3m5lddky4ak2r) about this today, a more bulletproof method can be found in the debug.ts file in the nuqs source code.

packages/docs/src/registry/items/use-date-range.json
```diff
@@ -0,0 +1,49 @@
+{
+  "type": "registry:item",
+  "name": "use-date-range",
```
franky47: **suggestion**: I've been wondering about registry item naming, particularly for parsers, where it may be better to keep the same name as the main exported thing (so `useDateRange` here). What do you think?

```
npx shadcn@latest add @nuqs/use-date-range
npx shadcn@latest add @nuqs/useDateRange
```

franky47 (commented):
This is awesome, thanks! 🙌

packages/docs/src/registry/items/use-date-range.json
iHiteshAgrawal: Agreed. `@nuqs/useDateRange` feels more natural. I'll update.

packages/docs/src/registry/items/useDateRange.json
```diff
@@ -0,0 +1,49 @@
+{
+  "type": "registry:item",
+  "name": "use-date-range",
+  "title": "Date Range Hook",
+  "description": "Date range selection with URL synchronization and preset support. Includes common presets like last 7 days, last 30 days, and custom ranges.",
+  "dependencies": ["nuqs", "date-fns@^4.1.0"],
+  "files": [
+    {
+      "type": "registry:file",
+      "path": "https://raw.githubusercontent.com/iHiteshAgrawal/nuqs-presets/main/registry/default/types/index.ts",
```
franky47: **suggestion**: I would pin these to a SHA-1 commit for immutability, then we can figure out an automation process for review + update when there's a change in your repo.

packages/docs/src/app/playground/debug-control.tsx
iHiteshAgrawal: The error actually occurred at build time (when running next build [here](https://github.com/47ng/nuqs/blob/82653f9d315eb27ad59918453e16013afb436325/packages/docs/package.json#L19)) during prerendering

iHiteshAgrawal:
resolved all. ready for review again.

franky47 (commented):
One issue I see with the way the files are laid out is that installing multiple hooks will overwrite each other (eg: `~/hooks/types/index.ts`), even though the contents are identical.

Overall, each hook should be a single file, and avoid installing multiple levels of barrel files which bring little value to the person adding it in their codebase.

We could also prefix them with the project name to avoid conflicts with possible existing userland hooks, eg `~/hooks/nuqs-presets/useFilters.ts`.

Anothing thing that makes nuqs' declarative model interesting is deriving other features from parser declarations (eg: loaders for server-side, serializer to generate href strings for links etc). I need to look into how it's structured in your repo, but having those exports would be good to allow composing those definitions in userland.

iHiteshAgrawal:
> One issue I see with the way the files are laid out is that installing multiple hooks will overwrite each other (eg: `~/hooks/types/index.ts`), even though the contents are identical.
> 
> Overall, each hook should be a single file, and avoid installing multiple levels of barrel files which bring little value to the person adding it in their codebase.
> 
> We could also prefix them with the project name to avoid conflicts with possible existing userland hooks, eg `~/hooks/nuqs-presets/useFilters.ts`.
> 
> Anothing thing that makes nuqs' declarative model interesting is deriving other features from parser declarations (eg: loaders for server-side, serializer to generate href strings for links etc). I need to look into how it's structured in your repo, but having those exports would be good to allow composing those definitions in userland.

Hey. I've updated the registry to address the feedback. Every hook is now a single file and self-contained.

PR #1217, opened 2025-11-15
fix: pass in full pathname + search in TSR to avoid trailing slashes

### LINKED ISSUE #1215: useQueryState causes a navigation to trailing slash with search param w/ TanStack Router
## Context

```
  "name": "nuqs-bug",
    "nuqs": "^2.7.3",
```

What framework are you using?

TanStack Router

```
    "@tanstack/react-devtools": "^0.7.0",
    "@tanstack/react-router": "^1.132.0",
    "@tanstack/react-router-devtools": "^1.132.0",
    "@tanstack/router-plugin": "^1.132.0",
```

## Description

When using useQueryState (or others) and calling the set(param) function, it causes a navigation, appending a trailing slash to the current route.

Expected behavior is that it doesn't cause a route transition with the path and only updates the query parameters.

## Reproduction

https://codesandbox.io/p/github/mirshko/nuqs-bug/main?import=true

you can test it easier on the the second page, click "To Second Page Test" in preview window.

franky47:
TBC if this method causes problems elsewhere (maybe running loaders when not needed?)

Also fixes persistence of custom `history.state` properties.

Closes #1215.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -26,6 +26,7 @@ When running `next dev`, this will:
   - http://localhost:3002 - [React SPA](./packages/e2e/react)
   - http://localhost:3003 - [Remix](./packages/e2e/remix)
   - http://localhost:3004 - [TanStack Router](./packages/e2e/tanstack-router)
+  - http://localhost:3005 - [React Router v5](./packages/e2e/react-router/v5)
   - http://localhost:3006 - [React Router v6](./packages/e2e/react-router/v6)
   - http://localhost:3007 - [React Router v7](./packages/e2e/react-router/v7)
 - Start the examples:
--- a/packages/docs/src/app/(pages)/_landing/quotes/quotes-section.tsx
+++ b/packages/docs/src/app/(pages)/_landing/quotes/quotes-section.tsx
@@ -296,7 +296,7 @@ export function QuotesSection() {
           name: 'Christopher Burns',
           handle: '@BurnedChris',
           avatar:
-            'https://pbs.twimg.com/profile_images/1967905434754334720/V6O4cTOV_400x400.jpg'
+            'https://pbs.twimg.com/profile_images/1988259821020221440/qRyYd6iE_400x400.jpg'
         }}
         text="Really glad to finally met Francois from nuqs! I love using it!"
         url="https://x.com/BurnedChris/status/1981071113154417018"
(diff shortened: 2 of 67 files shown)


packages/nuqs/src/adapters/tanstack-router.ts
```diff
@@ -56,20 +50,17 @@ function useNuqsTanstackRouterAdapter(watchKeys: string[]): AdapterInterface {
...
-          // a type error and possible basepath issues, so we switch it to '.' instead.
-          // See https://github.com/47ng/nuqs/pull/953#issuecomment-3003583471
-          to: renderQueryString(search) || '.',
-          // `from` will be handled by tanstack router match resolver, code snippet:
-          // https://github.com/TanStack/router/blob/5d940e2d8bdb12e213eede0abe8012855433ec4b/packages/react-router/src/link.tsx#L108-L112
-          ...(from ? { from } : {}),
+          // Note: we need to specify pathname + search here to avoid TSR appending
+          // a trailing slash to the pathname, see https://github.com/47ng/nuqs/issues/1215
+          from: '/',
+          to: location.pathname + renderQueryString(search),
```
SeanCassiere: You might want to pull the `pathname` out using `useLocation`. I'd use a selector in there, for fine-grained subscriptions.

franky47:
Thanks for your review @SeanCassiere (and good luck with the move!) 🫶

I doesn't work on the apex route indeed, and even this causes TSR to emit a trailing slash:
```ts
navigate({ href: 'https://example.com?foo=bar' })
// navigates to https://example.com/?foo=bar
```

I'll look into the TSR code to see if this can be fixed upstream, even if we're not following the intended use of the API, `trailingSlash` should be respected.

Edit: actually it doesn't seem possible to not have a trailing slash for the apex, even this adds one:

```ts
history.replaceState(null, '', 'https://example.com?foo=bar')
// navigates to https://example.com/?foo=bar
```

andreastande:
> Edit: actually it doesn't seem possible to not have a trailing slash for the apex, even this adds one:

Weird, do you think that is a bug with TSR? Should be possible

franky47:
> Weird, do you think that is a bug with TSR? Should be possible

No, this happens on https://example.com if you type `history.replaceState(null, '', 'https://example.com?foo=bar')`, so it's not related to any framework, it's a limitation of the web platform.

Actually even loading `https://example.com?foo=bar` in the browser gives you `https://example.com/?foo=bar`

andreastande:
TIL

PR #1254, opened 2025-12-05
chore: Vitest Browser Mode

ntatoud:
## Main goal
This PR aims to move away from JSdom for tests in a browser environment and use vitest browser mode instead.

## Additional notes

On top of moving tests needing a browser environment away from jsdom, I splitted existing tests in 3 vitest projects.

1. `browser` => Self explanatory (matches all test names `*.browser.test.*`)
For this, I had to replace the usage of some node apis in the tests since they are not polyfilled in the browser.

2. `unit` => For all tests that should pass in a non-browser environment. (matches all `*.test.*` except `*.browser.test.*`
3. `types` => For all type testings

- [x] Updated the CI to install necessary playwright browser.
- [x] Updated the `package.json` scripts to keep consistent behavior.

## Pending questions:
- Should we test against all browser available with `playwright` (Webkit, Firefox, Chromium) or keep only Chromium as it is right now ?

### DIFF
--- a/packages/nuqs/src/useQueryState.browser.test.tsx
+++ b/packages/nuqs/src/useQueryState.browser.test.tsx
@@ -1,7 +1,7 @@
 import React, { useState } from 'react'
 import { describe, expect, it, vi } from 'vitest'
+import { render, renderHook } from 'vitest-browser-react'
 import { page, userEvent } from 'vitest/browser'
-import { renderHook, render } from 'vitest-browser-react'
 import {
   NullDetector,
   useFakeLoadingState
@@ -283,19 +283,15 @@ describe('useQueryState: update sequencing', () => {
     let p0: Promise<URLSearchParams> | undefined = undefined
     let p1: Promise<URLSearchParams> | undefined = undefined
     let p2: Promise<URLSearchParams> | undefined = undefined
+    // prettier-ignore
     await act(async () => {
-      // prettier-ignore
       // Flush the queue from previous tests
       await new Promise(r => setTimeout(r, 60))
       // First, push an update to a to be emitted "immediately"
       p0 = result.current.a[1]('init')
       // Then two updates before the end of the throttle timeout
-      setTimeout(() => {
-        p1 = result.current.a[1]('a')
-      }, 10)
-      setTimeout(() => {
-        p2 = result.current.b[1]('b')
-      }, 20)
+      setTimeout(() => { p1 = result.current.a[1]('a') }, 10)
+      setTimeout(() => { p2 = result.current.b[1]('b') }, 20)
       return new Promise(resolve => setTimeout(resolve, 30))
     })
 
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-cd.yml
```diff
@@ -81,6 +81,8 @@ jobs:
           cache: pnpm
       - name: Install dependencies
         run: pnpm install --ignore-scripts --frozen-lockfile --filter nuqs...
+      - name: Install Playwright Chromium
+        run: pnpm exec playwright install chromium --with-deps
```
franky47: **note**: It looks like this step fails:

```
  pnpm exec playwright install chromium --with-deps
  shell: /usr/bin/bash -e {0}
  env:
    FORCE_COLOR: 3
    PNPM_HOME: /home/runner/setup-pnpm/node_modules/.bin
undefined
 ERR_PNPM_RECURSIVE_EXEC_FIRST_FAIL  Command "playwright" not found
Error: Process completed with exit code 254.
```

https://github.com/47ng/nuqs/actions/runs/19972510720/job/57284417778?pr=1254#step:6:9

Do we need to add playwright as a devDependency?

franky47 (commented):
Thank you!

We can test only in chromium for those tests, there are some things which may be useful to test in Safari (detecting the default throttle rate), but we can leave that out of scope for now.

.github/workflows/ci-cd.yml
ntatoud: Ah yes we do, I'ma fix that

ntatoud: I added it in the pnpm e2e catalog like cypress since you intend to migrate to it anyway.
I also removed jsdom that was still hanging there

franky47:
We have a different error now, on the [install step](https://github.com/47ng/nuqs/actions/runs/19975924996/job/57295517167?pr=1254#step:6:8):
```
None of the selected packages has a "bexec" script
```
And when the tests run (expectedly, it [doesn't find the binary](https://github.com/47ng/nuqs/actions/runs/19975924996/job/57295517167?pr=1254#step:7:164)):
```
Error: browserType.launch: Executable doesn't exist at /home/runner/.cache/ms-playwright/chromium_headless_shell-1200/chrome-linux/headless_shell
```

ntatoud:
> We have a different error now, on the [install step](https://github.com/47ng/nuqs/actions/runs/19975924996/job/57295517167?pr=1254#step:6:8):
None of the selected packages has a "bexec" script

Ah my bad, it should have been `exec` not bexec. Nice catch, it seems to work now

franky47:
It does, but it takes ~40s to install playwright 😭

I wonder if we could cache it, it's [not recommended](https://playwright.dev/docs/ci#caching-browsers) but surely restoring from cache would be faster (and more robust) than going through the network?

Do we also need system dependencies? GHA images contain a lot of things built-in.
Edit: looks like it comes down to 9s without the system deps, and tests still pass 🙌

franky47:
The build fails on Vercel but the error seems completely unrelated (unless it's a lockfile update cascading).

I'll review over the weekend, thanks again!

PR #1271, opened 2025-12-20
doc: add LLM features (copy as markdown, open button, llms.txt)

mezotv:
Adds a copy as markdown & open in x to the docs! Also bumped nextjs globally!

### DIFF
--- a/packages/docs/src/app/docs/[[...slug]]/page.tsx
+++ b/packages/docs/src/app/docs/[[...slug]]/page.tsx
@@ -8,7 +8,12 @@ import {
 } from '@/src/components/ai/page-actions'
 import { getBaseUrl } from '@/src/lib/url'
 import { github } from '@/src/lib/utils'
-import { DocsBody, DocsPage, DocsTitle } from 'fumadocs-ui/page'
+import {
+  DocsBody,
+  DocsDescription,
+  DocsPage,
+  DocsTitle
+} from 'fumadocs-ui/page'
 import type { Metadata } from 'next'
 import { notFound } from 'next/navigation'
 import { stat } from 'node:fs/promises'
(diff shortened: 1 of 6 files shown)


packages/docs/src/app/docs/[[...slug]]/page.tsx
```diff
@@ -36,7 +37,16 @@ export default async function Page(props: PageProps) {
       }}
     >
       <DocsTitle>{page.data.title}</DocsTitle>
-      <DocsDescription>{page.data.description}</DocsDescription>
+      <p className="text-lg text-fd-muted-foreground mb-2">
+        {page.data.description}
+      </p>
+      <div className="flex flex-row flex-wrap gap-2 items-center border-b pb-6">
+        <LLMCopyButton markdownUrl={`${page.url}.mdx`} />
```
Comment (franky47) [inline packages/docs/src/app/docs/[[...slug]]/page.tsx]: **issue**: This doesn't work for nested pages, eg: https://nuqs-6jlvxnimn-47ng.vercel.app/docs/parsers/built-in

packages/docs/src/app/docs/[[...slug]]/page.tsx
```diff
@@ -36,7 +37,16 @@ export default async function Page(props: PageProps) {
...
+        <ViewOptions
+          markdownUrl={`${page.url}.mdx`}
+          githubUrl={`https://github.com/${github.owner}/${github.repo}/blob/${github.branch}/packages/docs/content/docs/${page.path}`}
```
Comment (franky47) [inline packages/docs/src/app/docs/[[...slug]]/page.tsx:48]: **issue**: This produced a valid URL for "Open on GitHub", but invalid one (https://nuqs-6jlvxnimn-47ng.vercel.app/docs/parsers/built-in.mdx) for the other "Open in {AI chatbot}" options.

packages/docs/src/app/llms.mdx/[slug]/route.tsx
```diff
@@ -0,0 +1,34 @@
+import { type NextRequest, NextResponse } from 'next/server';
+import { getLLMText } from '@/src/lib/get-llm-text';
+import { source } from '@/src/app/source';
+import { notFound } from 'next/navigation';
+
+export const revalidate = false;
+
+export async function GET(
+  _req: NextRequest,
+  { params }: { params: Promise<{ slug: string }> },
```
Comment (franky47) [inline packages/docs/src/app/llms.mdx/[slug]/route.tsx]: **suggestion**: Type this using RouteContext, which should be type-safe as we use typedRoutes.

https://nextjs.org/docs/app/getting-started/route-handlers#route-context-helper

packages/docs/src/app/llms.mdx/[slug]/route.tsx
```diff
@@ -0,0 +1,34 @@
...
+) {
+  const { slug } = await params;
+  // Convert the slug string to an array (split by '/') for getPage()
+  // e.g., "parsers/built-in" -> ["parsers", "built-in"]
+  const slugArray = slug.split('/').filter(Boolean);
```
Comment (franky47) [inline packages/docs/src/app/llms.mdx/[slug]/route.tsx]: **note**: I think the issue here is that the slug should be a catch-all (`[...slug]`) to support nested doc paths. In which case it would already be an array.

packages/docs/content/docs/internal/design-system.mdx
```diff
@@ -1,6 +1,7 @@
 ---
 title: Design System
 description: Who needs Storybook anyway?
+exposeTo: [user]
```
franky47, line 4: **praise**: This API is 🔥

PR #1277, opened 2025-12-23
style(docs): Update title styles on pages and more responsive on mobile

necrydark:
Updated styling across all playground pages to make they are all follow the same styling, based on the first playground page of basic counter).

Made the pages more responsive on mobile by adding padding on smaller screens.

Love the project!

### DIFF
--- a/packages/docs/src/app/playground/(demos)/_components/source-on-github.tsx
+++ b/packages/docs/src/app/playground/(demos)/_components/source-on-github.tsx
@@ -18,7 +18,8 @@ export async function SourceOnGitHub({ path }: SourceOnGitHubProps) {
         </span>
         <a
           href={`https://github.com/47ng/nuqs/tree/next/packages/docs/src/app/playground/(demos)/${path}`}
-          className="ml-auto text-sm"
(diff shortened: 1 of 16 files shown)


franky47:
Thanks! The spacing looks much better indeed, but we could go further and use the same typography primitives for the title, description etc.

Normalising the code blocks would be nice too.

necrydark:
Thanks! I've added back the typography I removed and changes the other playground files to include missing typography where needed.

franky47:
Hum it doesn't look like the rest of the pages:

CleanShot 2025-12-24 at 07 15 32@2x

CleanShot 2025-12-24 at 07 16 02@2x

CleanShot 2025-12-24 at 07 16 22@2x

Feel free to use the registry pages as an example, those should have the right components & structure.

necrydark:
Thanks, I'll fix the styling to be the same as registry and other pages.

necrydark:
Screenshot 2025-12-24 162042
Screenshot 2025-12-24 162021

They should be the same now sorry for the hassle about it!

franky47:
seems to have been fixed by the update in #1493.

PR #1281, opened 2025-12-25
doc: make sure nextjs weekly switches color

mezotv:
Makes sure the nextjs weekly sponsor looks visible on light and dark mode!

Before:
image

After: 
image


franky47 (approved):

franky47:
Good catch, thanks!

PR #1284, opened 2025-12-26
doc: OSS Pledge details

franky47:
- Can't join as an individual member yet, so I'm signing up nuqs
- Add other contributions I made during the year
- Add a total summary


vladh:
LGTM! 😊

PR #1303, opened 2026-01-03
feat(adapters): add React Router v6 HashRouter support

### LINKED ISSUE #810: react-router HashRouter support?
## Context

```
-> https://pkg.pr.new/nuqs@807
```

What framework are you using?

- ❌ Next.js (app router)
- ❌ Next.js (pages router)
- ❌ React SPA (no router)
- ❌ Remix
- ✅ React Router
- ❌ Other (please specify)

```
-> 7.0.2
```

## Description

For some reason, we can't use BrowserRouter in RR7, only HashRouter allowed. But useQueryState always append the whole hash part.

Expect:
```
http://localhost:5173/#/form
⇓
http://localhost:5173/#/form?keyword=kw
⇓
http://localhost:5173/#/form?keyword=kw1
⇓
http://localhost:5173/#/form?keyword=kw12
⇓
http://localhost:5173/#/form?keyword=kw123
```

Actual:
```
http://localhost:5173/#/form
⇓
http://localhost:5173/#/form?keyword=kw#/form
⇓
http://localhost:5173/#/form?keyword=kw1#/form?keyword=kw#/form
⇓
http://localhost:5173/#/form?keyword=kw12#/form?keyword=kw1#/form?keyword=kw#/form
⇓
http://localhost:5173/#/form?keyword=kw123#/form?keyword=kw12#/form?keyword=kw1#/form?keyword=kw#/form
```

## Reproduction

https://github.com/Jungzl/react-router-nuqs-hash-issue
1. go to '/form'
2. type something, press Enter
3. see url

### LINKED ISSUE #1173: nuqs generates incorrect URL format with React Router's createHashRouter - query params before hash fragment
## Context
```
 "nuqs": "^2.7.1",
```
What framework are you using?
- ✅ React Router (v6)
- ✅ Using Single SPA (with hash routing)

**Note:** Tried using `NuqsAdapter` from `nuqs/adapters/react-router/v6` but it did not resolve the issue.

## Description

When using `nuqs` with React Router's `createHashRouter` and a `basename`, the generated URLs place query parameters **before** the hash fragment instead of after it.

### Expected Behavior
```
http://localhost:9000/#/test/data?aprovadores=9&status=pendente
```

### Actual Behavior
```
http://localhost:9000/?aprovadores=9&status=pendente#/test/data
```

### Environment
- **nuqs version**: [latest]
- **React Router version** (v6): 
- **Setup**: Single-spa with microfrontends using hash routing

### Steps to Reproduce

1. Create a hash router with basename:
```typescript
export const router = createHashRouter(
  [
    {
      path: '/data',
      element: <Aprovacoes />,
    },
  ],
  { basename: '/test' }
);
```

2. Use `useQueryStates` in a component:
```typescript
import { useQueryStates, parseAsArrayOf, parseAsString } from 'nuqs';

export function Aprovacoes() {
  const [filters, setFilters] = useQueryStates({
    aprovadores: parseAsArrayOf(parseAsString),
    status: parseAsString,
  });

  return <div>{/* component */}</div>;
}
```

3. Update query state with `setFilters({ aprovadores: ['9'] })`

### Result
Query params are placed before the hash instead of after.

### Additional Context
This is a microfrontend architecture using single-spa where hash routing is required. The `NuqsAdapter` from `nuqs/adapters/react-router/v6` does not resolve this issue.

gsaandy:
## Summary
  - Adds a dedicated adapter for React Router v6 HashRouter (`createHashRouter` and `<HashRouter>`)
  - Correctly handles the HashRouter URL structure where search params are inside the hash fragment
  (`/#/page?foo=bar` instead of `/?foo=bar`)
  - Includes full e2e test suite and documentation updates

  ## New Files
  - `nuqs/adapters/react-router/v6-hash` - Entry point
  - `src/adapters/lib/hash-router.ts` - Factory function
  - `src/adapters/lib/hash-router-utils.ts` - URL parsing utilities
  - `packages/e2e/react-router/v6-hash/` - E2E test suite

  ## Test plan
  - [x] Unit tests pass (`pnpm run test:unit`)
  - [x] Build succeeds (`pnpm run build`)
  - [x] Manually tested in a real app using `createHashRouter`

  Closes #810
  Closes #1173

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,6 +148,40 @@ export function ReactRouter() {
 
 </details>
 
+<details><summary><span style="width:16px;height:16px;background:#fff;border-radius:2px;"><img width="16px" height="16px" src="https://reactrouter.com/_brand/React%20Router%20Brand%20Assets/React%20Router%20Logo/Light.svg" /></span> React Router v6 (HashRouter)
+</summary>
+
+> Supported React Router versions: `react-router-dom@^6`
+>
+> For applications using `HashRouter` or `createHashRouter`, where the URL path and search params are stored in the hash fragment (e.g., `/#/page?foo=bar`).
+
+```tsx
+import { NuqsAdapter } from 'nuqs/adapters/react-router/v6-hash'
+import { createHashRouter, RouterProvider } from 'react-router-dom'
+import App from './App'
+
+const router = createHashRouter([
+  {
+    path: '/',
+    element: <App />
+  }
+])
+
+export function ReactRouter() {
+  return (
+    <NuqsAdapter>
+      <RouterProvider router={router} />
+    </NuqsAdapter>
+  )
+}
+```
+
+Note: HashRouter has some limitations compared to BrowserRouter:
+- Native HTML forms don't work (they submit to `location.search`, not the hash)
+- `shallow: false` has no effect (hash is never sent to server)
+
+</details>
+
 <details><summary><span style="width:16px;height:16px;background:#fff;border-radius:2px;"><img width="16px" height="16px" src="https://reactrouter.com/_brand/React%20Router%20Brand%20Assets/React%20Router%20Logo/Light.svg" /></span> React Router v7
 </summary>
 
--- a/packages/nuqs/src/adapters/lib/hash-router-utils.test.ts
+++ b/packages/nuqs/src/adapters/lib/hash-router-utils.test.ts
@@ -32,15 +32,21 @@ describe('hash-router-utils', () => {
     })
 
     it('handles multiple query params', () => {
-      expect(getSearchFromHash('#/page?foo=bar&baz=qux')).toBe('?foo=bar&baz=qux')
+      expect(getSearchFromHash('#/page?foo=bar&baz=qux')).toBe(
+        '?foo=bar&baz=qux'
+      )
     })
 
     it('handles nested paths', () => {
-      expect(getSearchFromHash('#/deep/nested/path?key=value')).toBe('?key=value')
+      expect(getSearchFromHash('#/deep/nested/path?key=value')).toBe(
+        '?key=value'
+      )
     })
 
     it('preserves everything after the first ?', () => {
-      expect(getSearchFromHash('#/page?foo=bar?baz=qux')).toBe('?foo=bar?baz=qux')
+      expect(getSearchFromHash('#/page?foo=bar?baz=qux')).toBe(
+        '?foo=bar?baz=qux'
+      )
     })
   })
 


franky47:
Wow, that's quite impressive, thanks!

Since the hash part is never sent to the server, I don't think the shallow: false option makes sense for this adapter, it's only ever going to be used client-side (since RRv6 doesn't do SSR). It's mostly a way to control whether to call the loader or not (some sort of inverted logic for `shouldRevalidate`, if this API exists in v6, I'm mostly familiar with it in v7).

Feel free to add a job to the ci-cd.yml so we can have the test results in here.

gsaandy:
> Wow, that's quite impressive, thanks!
> 
> Since the hash part is never sent to the server, I don't think the shallow: false option makes sense for this adapter, it's only ever going to be used client-side (since RRv6 doesn't do SSR). It's mostly a way to control whether to call the loader or not (some sort of inverted logic for `shouldRevalidate`, if this API exists in v6, I'm mostly familiar with it in v7).
> 
> Feel free to add a job to the ci-cd.yml so we can have the test results in here.

@franky47 - Updated the PR, could you take another look.

packages/e2e/react-router/v6-hash/package.json
```diff
@@ -0,0 +1,31 @@
+{
+  "name": "e2e-react-router-v6-hash",
+  "private": true,
+  "version": "0.0.0",
+  "type": "module",
+  "scripts": {
+    "dev": "vite --port 3016",
```
franky47, line 7: **praise**: Good way to indicate we're still on RRv6, I like it.

**suggestion**: This should be added to the list of ports used in the `CONTRIBUTING.md` document.

packages/nuqs/src/adapters/lib/hash-router-utils.ts
franky47, line 1: **issue**: These utility functions should be unit-tested against various (passing & edge) cases.

packages/nuqs/src/adapters/lib/hash-router.ts
franky47, line 1: **note**: It's hard to see the difference without an actual diff, but it looks like there is a lot of overlap between this implementation and that of the `location.search`-based adapters for RRv6/v7/Remix.

Would it make sense to make the generic implementation configurable to route to either the query string or the hash? (which may even open the possibility of having a hash-based behaviour for RRv7, not sure if Remix supports it).

packages/e2e/react-router/v6-hash/specs/hash-router.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+  const hashContent = hash.startsWith('#') ? hash.slice(1) : hash
+  const searchIndex = hashContent.indexOf('?')
+  return searchIndex >= 0 ? hashContent.slice(searchIndex) : ''
+}
+
+test.describe('HashRouter Basic I/O', () => {
+  test('reads the value from the URL on mount', async ({ page }) => {
+    await page.goto('./#/basic-io/useQueryState?test=init')
+    await page.waitForLoadState('networkidle')
+    await page.locator('#hydration-marker').waitFor({ state: 'hidden' })
```
franky47: **suggestion**: This could use the updated `navigateTo` implementation.

packages/e2e/react-router/v6-hash/specs/hash-router.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+ * e.g., "http://localhost:3016/#/page?foo=bar" => "?foo=bar"
+ */
+function getSearchFromHash(url: URL): string {
+  const hash = url.hash
+  const hashContent = hash.startsWith('#') ? hash.slice(1) : hash
+  const searchIndex = hashContent.indexOf('?')
+  return searchIndex >= 0 ? hashContent.slice(searchIndex) : ''
+}
+
+test.describe('HashRouter Basic I/O', () => {
```
franky47: **note**: Even though the URL structure is different than the BasicIO test, this should probably be placed in its own basic-io.spec.ts file to help sorting/filtering/maintaining test specs in the future.

packages/e2e/react-router/v6-hash/specs/shared.spec.ts
franky47, line 1: **issue**: Ah so that explains why the tests are passing, a lot of the test fixtures (routes) are actually not covered with specs.

They would actually not benefit from using the shared `*.spec.ts` files in e2e-shared, as those are querystring-based, but the rule of thumb is if a route is defined in the app, it should be covered by a test.

For an adapter to be officially supported in core, the list of tests defined in `runSharedTests` in `packages/e2e/shared/shared.spec.ts` is the minimum required. Some might actually not make sense in this case (eg: hash preservation, I don't think we can have another layer of hash at the end of the "query string" in the hash router). The other specs would need rewriting to handle the different setup & assertions on the URL (unless you have an idea on how to make it generic, using the abstractions you have in place for `navigateTo` and `.toHaveUrl` matchers. Note that we also have custom ones in `expect-url.ts`).

gsaandy:
> Thanks, I took a look and the main issue is the lack of actual test specs for the routes that are declared in the e2e-react-router-v6-hash package.
> 
> Once we actually have the basics (shared behaviour testing) covered, we could consider this a core adapter. If it works fine enough with manual testing, I'm also happy to add it as a community adapter as a shadcn registry item.

Thanks @franky47 for reviewing this, I ran it through claude and here is the refactoring plan, could you review this?

## Plan: Address Review Comments for React Router v6 HashRouter Adapter

### Review Comments Summary

1. **Main issue**: Lack of actual test specs for routes - shared tests are minimum required for core adapter
2. **Port documentation**: Add port 3016 to CONTRIBUTING.md
3. **Unit tests**: hash-router-utils.ts needs unit tests
4. **Code overlap**: Refactor to make configurable implementation for hash vs query string routing
5. **Use navigateTo helper**: hash-router.spec.ts should use shared navigateTo
6. **Split test files**: Organize tests into separate spec files
7. **Shared tests requirement**: `runSharedTests` is minimum for core adapter support

---

## Part 1: Refactor Adapters (Strategy Pattern)

### Goal
Merge `hash-router.ts` into `react-router.ts` using a configurable `mode` parameter ('standard' | 'hash').

### New File: `packages/nuqs/src/adapters/lib/router-strategies.ts`

Create strategy interface and implementations:

```typescript
type RouterMode = 'standard' | 'hash'

interface RouterModeStrategy {
  getSearchParams(): URLSearchParams
  getSearchString(): string
  constructUrl(search: URLSearchParams): URL
  navigationEvents: readonly ('popstate' | 'hashchange')[]
  supportsServerNavigation: boolean
  getQueueResetMutex(shallow: boolean): 1 | 2
  extractSearchParamsFromUrl(url: URL | string): URLSearchParams | null
  extractSearchStringFromUrl(url: URL | string): string | null
}
```

- `createStandardStrategy()` - Uses `location.search`, `popstate` only
- `createHashStrategy()` - Uses `location.hash`, `popstate` + `hashchange`

### Modify: `packages/nuqs/src/adapters/lib/react-router.ts`

Update factory function signature:

```typescript
type CreateReactRouterBasedAdapterArgs = {
  adapter: string
  useSearchParams: UseSearchParams
  mode?: RouterMode  // Defaults to 'standard'
  useNavigate?: UseNavigate  // Only required for 'standard' mode
}
```

Key changes:
- Accept `mode` parameter
- Use strategy for URL reading/construction
- Conditionally call `navigate()` based on `strategy.supportsServerNavigation`
- Subscribe to events from `strategy.navigationEvents`

### Modify: `packages/nuqs/src/adapters/lib/patch-history.ts`

Update `patchHistory` to accept strategy:

```typescript
export function patchHistory(
  emitter: Emitter<SearchParamsSyncEmitterEvents>,
  adapter: string,
  strategy: RouterModeStrategy
): void
```

- Use `strategy.getSearchString()` for tracking
- Subscribe to `strategy.navigationEvents`
- Use `strategy.extractSearchStringFromUrl()` in sync function

### Update Entry Points

**v6.ts**: Add `mode: 'standard'` (explicit)
**v6-hash.ts**: Use unified factory with `mode: 'hash'`
**v7.ts**: Add `mode: 'standard'` (explicit)

### Delete: `packages/nuqs/src/adapters/lib/hash-router.ts`

Merged into unified implementation.

---

## Part 2: Update Shared Test Infrastructure

### Modify: `packages/e2e/shared/playwright/expect-url.ts`

Add `isHashRouter` parameter:

```typescript
export function expectSearch(
  page: Page,
  expected: Record<string, string>,
  isHashRouter = false
) {
  return expectUrl(page, url => {
    const searchString = isHashRouter
      ? getSearchFromUrl(url, true)
      : url.search
    const params = new URLSearchParams(searchString)
    return Object.entries(expected).every(
      ([key, value]) => params.get(key) === value
    )
  }, ...)
}
```

### Modify: All Shared Specs

Update URL assertions to use `config.isHashRouter`:

```typescript
// Before:
await expect(page).toHaveURL(url => url.search === '?test=pass')

// After:
await expect(page).toHaveURL(
  createSearchMatcher('?test=pass', config.isHashRouter ?? false)
)
```

Pass `isHashRouter` to `navigateTo`:
```typescript
await navigateTo(page, path, '?test=init', { isHashRouter: config.isHashRouter })
```

**Specs to update**:
- basic-io.spec.ts
- push.spec.ts
- json.spec.ts
- key-isolation.spec.ts
- shallow.spec.ts
- stitching.spec.ts
- native-array.spec.ts
- debounce.spec.ts
- flush-after-navigate.spec.ts
- repro-359.spec.ts
- repro-982.spec.ts
- repro-1099.spec.ts

### Modify: `packages/e2e/shared/shared.spec.ts`

Skip `hash-preservation` for HashRouter (can't have hash within hash):

```typescript
if (!config.isHashRouter) {
  testHashPreservation({ path: `${pathPrefix}/hash-preservation`, ...config })
}
```

---

## Part 3: Enable Shared Tests for v6-hash

### Modify: `packages/e2e/react-router/v6-hash/specs/shared.spec.ts`

```typescript
import { runSharedTests } from 'e2e-shared/shared.spec.ts'

runSharedTests('', {
  router: 'react-router-v6-hash',
  isHashRouter: true
})
```

### Modify: `packages/e2e/react-router/v6-hash/specs/hash-router.spec.ts`

- Keep only HashRouter-specific URL structure tests
- Remove Basic I/O tests (now covered by shared tests)
- Use `navigateTo` helper instead of inline `page.goto`

---

## Part 4: Unit Tests for hash-router-utils.ts

### New File: `packages/nuqs/src/adapters/lib/hash-router-utils.test.ts`

Test cases:

**getSearchFromHash**:
- `'#/page?foo=bar'` → `'?foo=bar'`
- `'#/page'` → `''`
- `'#?foo=bar'` → `'?foo=bar'`
- `''` → `''`
- `'#'` → `''`
- `'#/page?foo=bar?baz=qux'` → `'?foo=bar?baz=qux'` (first ? starts search)

**getPathnameFromHash**:
- `'#/page?foo=bar'` → `'/page'`
- `'#/page'` → `'/page'`
- `'#/'` → `'/'`
- `'#?foo=bar'` → `''`
- `''` → `''`

**constructHash**:
- `('/page', '?foo=bar')` → `'#/page?foo=bar'`
- `('/page', '')` → `'#/page'`
- `('', '?foo=bar')` → `'#?foo=bar'`

---

## Part 5: Documentation

### Modify: `CONTRIBUTING.md`

Add to port table:
```
| 3016 | React Router v6 (HashRouter) | `./packages/e2e/react-router/v6-hash` |
```

---

gsaandy:
@franky47 please review the new changes, and let me know if this approach works or not.

gsaandy:
@franky47 gentle reminder

PR #1316, opened 2026-01-30
fix: wait for navigation in RR adapters before syncing

### LINKED ISSUE #947: Remix - search params should not update before navigation completed
## Context

```
    "nuqs": "2.4.1",
```

What framework are you using?

- ✅ Remix

```
    "@remix-run/react": "^2.9.2",
```

## Description

Search params are updated before navigation completed which causes irrelevant renders and effects which should not happen.

Expected behaviour - search params should update while navigation not complete.

## Reproduction

Imagine this case:
- current search param `?category=all`
- click on navigation link to somewhere, zero search params, page not rendered yet
- - 🔴 bug here because old page still rendered, and search params from `nuqs` are already updated to their empty or default values
- navigation complete, page rendered

```tsx
  // use simple param
  const [immediateCategory, setCategory] = useQueryState(
    'category',
    parseAsString.withDefault('all'),
  );

  console.log(immediateCategory); // 🔴 see value resets to default when navigation started and page still mounted

  // step one - switch to `?category=red`
  // click link to navigate somewhere
  // 🔴 BUG - SomeLargeLayoutShiftComponent rendered because params updated to default value even though user did not navigate to "all" category, he just navigated to other page that has no search params at all
  return <div>
    {immediateCategory === 'all' && <SomeLargeLayoutShiftComponent />}
    <Link to="/some-page">some page</Link>
  </div>;
```

franky47:
See discussion #1293, fix by @itay-perry 🫶

Fixes #947.


itay-perry:
This is amazing @franky47 thank you so much!
When will it be released? :pray:

PR #1323, opened 2026-02-02
fix: stale query state with App Router cached components

### LINKED ISSUE #1273: nuqs state not updating correctly when using Next.js `cacheComponents=true`
### Discussed in https://github.com/47ng/nuqs/discussions/1272

<sup>Originally posted by **rajput-hemant** December 22, 2025</sup>
Hi 👋
I’m running into an issue with **nuqs** where the URL state seems to stay **stale** when `cacheComponents: true` is enabled in `next.config.js`.

### Context

I have a task list → task edit flow where:

* The selected `taskId` is stored in the URL using **nuqs**
* When clicking **Edit**, I navigate to a task form page
* The form is prefilled based on the `taskId` from nuqs state

This works fine **until a previously edited task is deleted**.

### Problem

When `cacheComponents: true` is enabled:

* nuqs continues to return the **previous taskId**
* Even after navigating back to the list and selecting a different task
* This causes the app to try fetching a **deleted task**, resulting in an error

This behavior **does not occur** when `cacheComponents` is disabled.

I’m not sure if this is:

* A bug in nuqs
* An interaction issue with Next.js component caching
* Or expected behavior, and I need to explicitly opt out of caching nuqs state

---

## **Expected behavior**

* nuqs state should reflect the **latest URL param**
* When navigating back and selecting a new task, the old `taskId` should not be reused
* Deleting a task should not cause stale URL state to persist

---

## **Actual behavior**

* nuqs returns a **stale `taskId`**
* App attempts to fetch a task that no longer exists
* This only happens when `cacheComponents: true` is enabled

---

## **Steps to reproduce**

1. Enable `cacheComponents: true` in `next.config.js`
2. Open the task list page
3. Click **Edit** on **Task A**

   * You are navigated to the edit form
   * Form is prefilled correctly
   * nuqs state contains `taskId = A`
4. Navigate back to the task list
5. Delete **Task A**
6. Click **Edit** on **Task B**
7. ❌ App still reads `taskId = A` from nuqs
8. ❌ Fetch fails because Task A no longer exists

---

## **Reproduction sandbox**

I’ve created a minimal CodeSandbox with everything needed to reproduce the issue:

👉 [https://codesandbox.io/p/devbox/32lkt4](https://codesandbox.io/p/devbox/32lkt4)

---

## **Questions**

1. Is this behavior expected when using `cacheComponents: true`?
2. If yes, what’s the recommended way to:

   * Opt out of this caching for nuqs state?
   * Or force nuqs to always sync with the latest URL params?
3. If not expected, I’m happy to raise an issue.

Thanks in advance 🙏
Appreciate all the work on **nuqs** — it’s been great otherwise!

https://github.com/user-attachments/assets/637f8136-421b-433d-b1dd-06c5869d59d6

</div>

AyushNigam-1:
### Problem

When `experimental.cacheComponents` is enabled in the Next.js App Router,
client components may be reused across navigations. `useQueryStates`
previously read search params at render time, which can become stale in
this scenario and cause outdated query state to persist.

This reproduces the issue described in #1273.

### Solution

This change removes render-time dependency on search params and instead
synchronizes query state reactively using the adapter snapshot API. This
ensures query state stays in sync with the URL even when components are
cached, without changing the public API.

### Notes

- No public API changes
- Behavior unchanged for non-cached routes
- Fix scoped to App Router caching behavior

Closes #1273

### DIFF
--- a/packages/nuqs/src/useQueryStates.ts
+++ b/packages/nuqs/src/useQueryStates.ts
@@ -37,8 +37,8 @@ export type Values<T extends UseQueryStatesKeysMap> = {
   [K in keyof T]: T[K]['defaultValue'] extends NonNullable<
     ReturnType<T[K]['parse']>
   >
-    ? NonNullable<ReturnType<T[K]['parse']>>
-    : ReturnType<T[K]['parse']> | null
+  ? NonNullable<ReturnType<T[K]['parse']>>
+  : ReturnType<T[K]['parse']> | null
 }
 type NullableValues<T extends UseQueryStatesKeysMap> = Nullable<Values<T>>
 
@@ -98,7 +98,6 @@ export function useQueryStates<KeyMap extends UseQueryStatesKeysMap>(
     [stateKeys, JSON.stringify(urlKeys)]
   )
   const adapter = useAdapter(Object.values(resolvedUrlKeys))
-  const initialSearchParams = adapter.searchParams
   const queryRef = useRef<Record<string, Query | null>>({})
   const defaultValues = useMemo(
     () =>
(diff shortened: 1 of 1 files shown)


franky47:
Closed in favour of #1320.

PR #1344, opened 2026-02-27
fix: use process.env.DEBUG for server-side debug detection

### LINKED ISSUE #1336: Node 25 warning on import: `--localstorage-file` was provided without a valid path
## Context

```bash
nuqs@2.8.8
```

What framework are you using?

- ✅ Next.js (app router)
- ❌ Next.js (pages router)
- ❌ React SPA (no router)
- ❌ Remix
- ❌ React Router
- ❌ Other

```bash
next@16.1.6
react@19.2.4
```

Additional environment:

```bash
node v25.6.0
macOS (arm64)
```

## Description

Importing `nuqs/server` on Node 25 emits this warning:

```text
(node:XXXX) Warning: `--localstorage-file` was provided without a valid path
    at Object.get (node:internal/webstorage:32:25)
```

In my app this appears during `next dev` (Turbopack) because server code imports `nuqs/server` for `createSearchParamsCache`.

From tracing, this happens at module evaluation time in `nuqs` debug code (`isDebugEnabled`) that probes `localStorage`.

## Reproduction

Minimal Node-only repro (no Next required):

```bash
mkdir nuqs-node25-repro && cd nuqs-node25-repro
npm init -y
npm i nuqs@2.8.8
NODE_OPTIONS=--trace-warnings node -e "import('nuqs/server').then(() => console.log('loaded'))"
```

Observed output includes:

```text
loaded
(node:...) Warning: `--localstorage-file` was provided without a valid path
    at Object.get (node:internal/webstorage:32:25)
    at isDebugEnabled (node_modules/nuqs/dist/server.js:...)
```

Expected:

- Importing `nuqs/server` in a server/runtime context should not touch `localStorage` at module top-level.
- No Node webstorage warning should be emitted.

Potential direction:

- Gate debug-localStorage checks behind `typeof window !== 'undefined'` (or defer the check to client-only paths).

ryannguyen95:
### Motivation:
fix: gate debug localStorage check so server import doesn’t trigger Node 25 warning

### Changes:
Use process.env.DEBUG when typeof window === 'undefined' and only read localStorage in the browser. Added tests and docs for server-side debugging.

Closes #1336.

### DIFF
--- a/packages/docs/content/docs/debugging.mdx
+++ b/packages/docs/content/docs/debugging.mdx
@@ -13,16 +13,18 @@ to `nuqs`, and reload the page.
 localStorage.setItem('debug', 'nuqs')
 ```
 
-Log lines will be prefixed with `[nuqs]` for `useQueryState` and `[nuq+]` for
-`useQueryStates`, along with other internal debug logs.
+Log lines for both `useQueryState` and `useQueryStates` will be prefixed with
+`[nuq+]`, along with other internal debug logs.
 
 > Note: unlike the `debug` package, this will not work with wildcards, but
 > you can combine it: `localStorage.setItem('debug', '*,nuqs')`
 
 ## Server (Node.js)
 
-In server or Node environments (e.g. when using `nuqs/server`), enable debug
-logs by setting the `DEBUG` environment variable so it contains `nuqs`:
+Debug logs apply in any Node environment: server-side rendering (SSR), React
+Server Components (RSC), or when using `nuqs/server`. Hooks like
+`useQueryState` and `useQueryStates` can run on the server in these contexts.
+Enable logs by setting the `DEBUG` environment variable so it contains `nuqs`:
 
 ```bash
 DEBUG=nuqs node server.js
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -82,10 +82,7 @@ In server or Node environments (e.g. when using `nuqs/server`), set the `DEBUG` 
 DEBUG=nuqs pnpm dev
 ```
 
-Log prefixes:
-
-- `[nuqs]` — single-key operations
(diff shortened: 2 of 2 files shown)


packages/docs/content/docs/debugging.mdx
```diff
@@ -11,11 +13,32 @@ to `nuqs`, and reload the page.
 localStorage.setItem('debug', 'nuqs')
 ```
 
+Log lines will be prefixed with `[nuqs]` for `useQueryState` and `[nuq+]` for
+`useQueryStates`, along with other internal debug logs.
```
franky47: **nitpick**: actually, since useQueryState is now just a thin wrapper around useQueryStates, it always shows `[nuq+]`, which for some weird compression reason yields a smaller gzipped bundle than `[nuqs]`. 🤷‍♂️

packages/docs/content/docs/debugging.mdx
```diff
@@ -11,11 +13,32 @@ to `nuqs`, and reload the page.
...
+
 > Note: unlike the `debug` package, this will not work with wildcards, but
 > you can combine it: `localStorage.setItem('debug', '*,nuqs')`
 
-Log lines will be prefixed with `[nuqs]` for `useQueryState` and `[nuq+]` for
-`useQueryStates`, along with other internal debug logs.
+## Server (Node.js)
+
+In server or Node environments (e.g. when using `nuqs/server`), enable debug
```
franky47: > (eg: when using `nuqs/server`)

**nitpick**: Not only, hooks can be server-side rendered too in RSC context (client components do also render on the server).

ryannguyen95: Ahh you’re right, I’ve just pushed a new commit to update it

PR #1348, opened 2026-03-02
doc: add registry parseAsColor using tinycolor2

jo-chemla:
See discussion https://github.com/47ng/nuqs/discussions/1342 Adding `parseAsColor` registry item here https://github.com/47ng/nuqs/tree/next/packages/docs/src/registry/items tinycolor is 5kB minified + gzipped, smaller than chromajs 16kB

### DIFF
--- a/packages/docs/src/registry/items/parser-color.source
+++ b/packages/docs/src/registry/items/parser-color.source
@@ -54,8 +54,10 @@ export function parseAsColor(options: ColorParserOptions = {}) {
 
   return createParser({
     parse(value: string) {
-      const color = tinycolor(value)
-      
+    // Decode URL-encoded characters (e.g., %23 -> #)
+    const decodedValue = decodeURIComponent(value)
+    const color = tinycolor(decodedValue)
+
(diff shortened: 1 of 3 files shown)


packages/docs/src/registry/items/parser-color.source
```diff
@@ -54,8 +54,10 @@ export function parseAsColor(options: ColorParserOptions = {}) {
 
   return createParser({
     parse(value: string) {
-      const color = tinycolor(value)
-      
+    // Decode URL-encoded characters (e.g., %23 -> #)
+    const decodedValue = decodeURIComponent(value)
```
franky47, line 58: **suggestion**: to keep the URL small we could forego the hash, just with the character set and its length we should be able to detect hex color codes, no?

jo-chemla: yes indeed, by default when the hash is not present the color is still correctly parsed. When serializing, we don't prefix with the hash, only use the 6 digits hex string.

PR #1355, opened 2026-03-09
feat!: add null to return type of serialize()

CatBraaain:
Allow the `serialize()` function to return `null` to support clearing the query parameter, similar to `useQueryState`.

### Background
In `useQueryState`, `null` consistently represents the absence of a query parameter:
- When reading, the returned value is `null` if the key is not present in the URL.
- When writing, passing `null` to the setter removes the key from the query string.

ref: 
- https://nuqs.dev/docs/basic-usage
- https://nuqs.dev/docs/batching#options

### Example

This change enables parsers to conditionally omit values from the URL. For example, a parser could return `null` for certain values to keep 
the URL clean:

```ts
const parser = createParser({
  parse: (value) => parseInt(value),
  serialize: (value) => value === 0 ? null : String(value)
})
```

When 0 is set, the parameter will be cleared from the URL instead of showing ?key=0.

### DIFF
--- a/packages/docs/content/docs/parsers/demos.tsx
+++ b/packages/docs/content/docs/parsers/demos.tsx
@@ -564,9 +564,11 @@ export function CustomMultiParserDemo() {
         return Object.entries(values)
           .map(([key, value]) => {
             if (!itemParser.serialize) return null
+            const serialized = itemParser.serialize(value)
+            if (serialized === null) return null
             return parseAsKeyValue.serialize({
               key,
-              value: itemParser.serialize(value)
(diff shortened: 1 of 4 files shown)


franky47:
This seems like a useful addition, but the widening of the return type for `serialize` is a breaking change (see the failure in the docs build).

I'll keep this one here for the v3 release line.

CatBraaain:
Thanks for taking a look.
I see about the breaking change, that makes sense.

I've pushed new commit fixes the type errors and formatting issues.

Issue #1245, opened 2025-12-03
State desynchronization when rendering <Navigate /> on first render


kouak:
## Context

What's your version of `nuqs`?

```
    "nuqs": "2.8.2",
```

What framework are you using?

- ❌ Next.js (app router)
- ❌ Next.js (pages router)
- ✅ React SPA (no router)
- ❌ Remix
- ✅ React Router

Which version of your framework are you using?

```
    "react-router": "^6.30.2",
    "react-router-dom": "^6.30.2",

```

## Description

When rendering a `<Navigate />` component which updates the search params on first render, `nuqs` state will never synchronize its state back.

## Reproduction

```typescript
import { render } from '@testing-library/react';
import userEvent from '@testing-library/user-event';
import { parseAsString, useQueryStates } from 'nuqs';
import { NuqsAdapter } from 'nuqs/adapters/react-router/v6';
import { BrowserRouter, Link, Navigate } from 'react-router-dom';
import { describe, expect, test } from 'vitest';

function TestComponent() {
  const [nuqsState] = useQueryStates({
    search: parseAsString.withDefault(''),
  });

  if (nuqsState.search === 'REDIRECT') {
    return (
      <Navigate
        to={{
          search: '?search=foo',
        }}
      />
    );
  }

  return (
    <>
      <span>{nuqsState.search}</span>
      <Link to={{ search: '?search=REDIRECT' }}>Trigger redirect</Link>
    </>
  );
}

describe('nuqs repro', () => {
  test('with initial search param', async () => {
    using _ = withInitialUrl('/?search=foo');

    const { container } = render(
      <NuqsAdapter>
        <BrowserRouter>
          <TestComponent />
        </BrowserRouter>
      </NuqsAdapter>,
    );

    expect(container).toHaveTextContent('foo');
  });

  test('with a click on a link which triggers the rendering of <Navigate />', async () => {
    using _ = withInitialUrl('/?search=foo');
    const user = userEvent.setup();

    const { container, getByRole } = render(
      <NuqsAdapter>
        <BrowserRouter>
          <TestComponent />
        </BrowserRouter>
      </NuqsAdapter>,
    );

    const redirectButton = getByRole('link', { name: 'Trigger redirect' });
    await user.click(redirectButton);

    expect(container).toHaveTextContent('foo');
  });

  test('with initial search param', async () => {
    /**
     * With this URL, <TestComponent /> will render <Navigate /> on first render.
     */
[...]

Issue #1156, opened 2025-10-02
Breaking changes between 2.6.0 and 2.7.0


l0gicgate:
## Context

What's your version of `nuqs`?

```
2.7.0
```

What framework are you using?

- ✅ Next.js (app router)

Which version of your framework are you using?
```
Next v15.5.4
```

## Description

When transitioning between routes, nuqs now holds previous route query parameters.

## Reproduction

1. Navigate to route 1 (using NextLink), set some query parameters
2. Navigate to route 2 (using NextLink), the query parameters from route 1 will still be there

This does not occur on v2.6.0

franky47:
Thanks for the report, I'm looking into it.

franky47:
I can't reproduce it with this setup:

```tsx
'use client'

import { useQueryState } from 'nuqs'
import { Display } from '../components/display'
import Link from 'next/link'

type Repro1156PageAProps = {
  pageBPath: string
}

export function Repro1156PageA({ pageBPath }: Repro1156PageAProps) {
  const [state, setState] = useQueryState('test')
  return (
    <>
      <button onClick={() => setState('pass')}>Set state</button>
      <Link href={pageBPath}>Navigate</Link>
      <Display environment="client" state={state} />
    </>
  )
}

export function Repro1156PageB() {
  const [state] = useQueryState('test')
  return <Display environment="client" state={state} />
}
```

Is there anything else you do in your code that might be relevant?

l0gicgate:
Here is a repo with the repro and here's a screenshot of the behavior:

https://github.com/l0gicgate/nuqs-issue-1156

# To reproduce
1. `bun i && bun run dev`
2. Open http://localhost:3000
3. Open console
4. Click on "Set query"
5. Then click on "Go to Page 2"
6. Look at the console output

`state1` from `page 1` should never bleed into `state2` on `page 2`

<img width="1285" height="674" alt="Image" src="https://github.com/user-attachments/assets/65b1a931-8566-4078-ad27-1c2ced2b9036" />

franky47:
Thanks for the reproduction, I see the same behaviour on nuqs@2.6.0 though.

Next.js renders the destination page in the background on navigation, and since the useQueryState(s) hooks see the URL as the source of truth (it hasn't changed yet while navigation is pending), they'd use the old state temporarily, until the URL updates to the new page, and the state follows that (returning null).

IIRC React Router & Remix also behave like this and there is an issue about this behaviour: #947.

[...]

PR #1359, opened 2026-03-12
test: add e2e reproduction for #1358

### LINKED ISSUE #1358: update another component while rendering a different component & unexpected query state
## Context

```
"nuqs": "^2.8.9",
```

What framework are you using?

- ✅ React Router

```
"react-router": "7.13.1"
```

## Description

1. React warning:
```
Cannot update a component (`RouteB`) while rendering a different component (`RouteA`). To locate the bad setState() call inside `RouteA`
```

2. The initial state was unexpected when navigating from another route. I conducted a simple investigation and found that the difference arises from React Router's internal search state and direct access to `location.search`.

## Reproduction

1. Go to https://codesandbox.io/p/sandbox/fz7g9k?file=%2Fsrc%2FRouteB.js
2. Switch between RouteA/RouteB
3. The input value was disordered

franky47:
## Summary

Fix cross-route state bleeding with setState during render for React Router & TanStack Router.

- [ ] Refactor to a cleaner implementation
- [ ] ~Split TanStack Router in separate PR?~
- [ ] Test against the warning in dev, not just the incorrect state update

Closes #1358.


franky47:
## Analysis of #1358 root cause

### The bug flow

When Route A (default `AAA`) navigates to Route B (default `BBB`), here's what happens:

1. **Route A renders**, `useQueryState('mode')` returns `null` → `if (!mode) { setMode('default') }` fires
2. The `update` function in `useQueryStates.ts:306` does: `emitter.emit(urlKey, { state: value, query })` — this is the **cross-hook sync emitter** (global singleton in `lib/sync.ts`)
3. Then it pushes to `globalThrottleQueue` and calls `.flush()`, which schedules the actual URL update via `setTimeout(0)` (next tick)
4. **User clicks link** → React Router navigation starts
5. **The queued flush fires** — `applyPendingUpdates()` in `throttle.ts:178` calls `getSearchParamsSnapshot()` which reads `new URLSearchParams(location.search)`, applies the pending `mode=default`, and calls `adapter.updateUrl()` which does `history.replaceState()` with `mode=default` in the URL
6. **Route B mounts** — its `useOptimisticSearchParams()` initializes from `location.search` (line 116 in `react-router.ts`), which now **includes `mode=default`** from Route A's flush
7. Route B's `useQueryState('filter', parseAsString.withDefault('BBB'))` — the `filter` key is not in the URL, but `mode=default` **is**, showing state from Route A leaked into Route B's URL

### Root cause: `autoResetQueueOnUpdate: false`

The React Router adapter (`adapters/lib/react-router.ts:94`) sets `autoResetQueueOnUpdate: false`. This means the `globalThrottleQueue` is **never cleared** between route navigations for the React Router adapter.

Compare with the **Next.js App Router** adapter (`adapters/next/impl.app.ts:103`) which sets `autoResetQueueOnUpdate: true` — this was the fix for the **identical** issue #1156 ("On navigation, Next.js renders the target page with the source URL state"), landed in PR #1320.

From the PR #1320 description:
> "the issue seemed to be that the queued queries were not cleared after a nuqs state update, until after the navigation transition had completed (which caused this pre-rendering of the destination page with the old queued queries, taking precedence over the correct values returned by useSearchParams)"

### Two additional contributing factors

**1. `location.search` vs React Router state** — The adapter reads `window.location.search` directly (lines 106, 116, 125) rather than using React Router's internal location state. React Router may have already transitioned its internal state to Route B's URL, but `window.location.search` still reflects Route A's pending flush, creating a window where old state bleeds into new components.

**2. Global singleton queues** — Both `globalThrottleQueue` (throttle.ts:221) and `debounceController` (debounce.ts:172) are module-level singletons. There's no scoping by route or component lifecycle. When Route A queues an update and navigation happens before the flush, the pending update persists and can execute in Route B's context.

### Related prior art

| Issue | Framework | Root cause | Fix |
|---|---|---|---|
| **#1156** | Next.js App Router | Queued queries not cleared after navigation | Set `autoResetQueueOnUpdate: true` (PR #1320) |
| **#947** | React Router / Remix | Search params updated before navigation completed | Wrap state updates in `startTransition()` (PR #1316) |
| **#1293** | React Router | Cross-page state bleeding on back/forward | Same `startTransition` fix |
| **#524** | Next.js Pages | Both source and target pages re-rendered during navigation | Pages router keeps both mounted during transition |
| **#1245** | React Router v6 | `<Navigate />` updates search params during first render, never syncs | Open — same pattern as #1358 |
| **#1099** | Next.js / React Router | State temporarily flashes old value | Queue reset timing fix |

### Suggested investigation path

The most direct avenue is whether setting `autoResetQueueOnUpdate: true` in the React Router adapter (like was done for Next.js App Router in PR #1320) resolves #1358 without breaking the existing flush-after-navigate tests. The flush-after-navigate tests specifically exercise debounce/throttle queue behavior across navigations, so they'd be the canary for regressions.

The cross-hook sync emitter firing synchronously during render (line 306) is the deeper issue causing the React warning, but addressing the queue persistence would at least prevent the state bleeding.

franky47:
Superseded by #1450 (simpler fix).

PR #1382, opened 2026-04-06
feat: add Storybook adapter for @storybook/nextjs-vite

shawngustaw:
Add a new `nuqs/adapters/storybook` adapter that works within Storybook's mock Next.js environment. The adapter reads initial search params from Storybook's mock `useSearchParams()` context, then manages state internally so nuqs updates reflect in the UI.

Exports `NuqsStorybookAdapter` component and `withNuqsStorybookAdapter()` decorator factory.

### DIFF
--- a/packages/nuqs/adapters/storybook.d.ts
+++ b/packages/nuqs/adapters/storybook.d.ts
@@ -1,0 +1,7 @@
+// This file is needed for projects that have `moduleResolution` set to `node`
+// in their tsconfig.json to be able to `import {} from 'nuqs/adapters/storybook'`.
+// Other module resolutions strategies will look for the `exports` in `package.json`,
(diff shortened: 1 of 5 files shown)


franky47:
Thanks for the PR!

Official adapters need an end-to-end test suite to be accepted in the core `nuqs` package. The easiest way to contribute this Storybook adapter would be to add it as a shadcn registry item (see how waku, inertia etc are done).

But if you want to add the full test suite, that'd be cool too, just warning you that it's a lot more work.

shawngustaw:
@franky47 I actually ended up getting this working using only the public api so published a package here (https://github.com/shawngustaw/nuqs-storybook-adapter) since I may discover some hidden bugs along the way, but feel free to point people there if this comes up again 🙏

PR #1392, opened 2026-04-29
perf: reduce nuqs bundle size by 440 bytes (-3.24%)

franky47:
## Why

Hard \`size-limit\` caps and "minimal bundle size" as a stated goal. Slack against the caps was effectively gone (Client 6 / 6000, minimal tree-shaken 25 / 4500, Server 398 / 3500), blocking new features without bumping limits.

After: 197, 139, 533. Total -440 B / -3.24%.

## Constraints respected

- Public API unchanged (\`api.test.ts\` runtime snapshot + \`tests/*.test-d.ts\`).
- Behavior unchanged (full nuqs suite + e2e-next 513/513).
- No tests, no \`size-limit\` thresholds, no build config touched.
- Every \`console.*\` / \`debug()\` / \`performance.mark\` call site preserved (count: 41 in src).

## Provenance

Each commit is one autonomous evo experiment; the trailer line ties it back to its experiment ID for spelunking. The trailing \`style:\` commit is just \`prettier --write\` after the cherry-pick.

## Test plan
- [x] \`pnpm test --filter nuqs\` — green
- [x] \`pnpm test --filter e2e-next\` — 513/513 green
- [ ] Smoke test in a downstream consumer
- [ ] Reviewer pass on short helper names (\`pe\`, \`ok\`)

PR #1394, opened 2026-04-29
chore: fix npm pkg set command

franky47:
Bumping the pnpm action setup step in #1392 bundles pnpm v11.0.0-rc.5, which doesn't forward unimplemented commands to npm (like npm set). For other commands, it uses the pinned version in package.json in packageManager.

Waiting until pnpm ships 11 GA in their pnpm/setup-action to upgrade to pnpm 11 everywhere.

### DIFF
--- a/.github/workflows/pkg.pr.new.yml
+++ b/.github/workflows/pkg.pr.new.yml
@@ -1,9 +1,9 @@
 name: PR Preview
 on:
   pull_request:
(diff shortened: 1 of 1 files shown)


franky47:
Looks like the pinned pnpm version has been fixed in https://github.com/pnpm/action-setup/pull/249

PR #1406, opened 2026-05-20
doc: add brand assets context menu

TheOrcDev:
## Summary
- add a Radix/shadcn context menu to copy nuqs brand SVG assets from the docs wordmark
- add a separated download action that zips all SVGs from packages/res in-browser
- wire sonner toasts without adding next-themes directly

## Test
- pnpm --filter docs test


TheOrcDev:
hm I see I have `❌ Version Bump Check Failed` 

is it needed for docs too @franky47 ?

franky47:
Yeah the feat/fix prefixes only apply to library code and control the version bump by semantic-release. Everything else is a doc/chore/etc.

packages/docs/src/components/brand-assets-menu.client.tsx
```diff
@@ -0,0 +1,179 @@
...
+
+export function BrandAssetsMenuClient({
+  assets,
+  downloadableAssets,
+  children
+}: BrandAssetsMenuClientProps) {
+  const copyAsset = async (asset: BrandAsset) => {
+    const copied = await copyText(asset.svg)
+    if (copied) {
+      toast.success(`${asset.toastLabel} copied`)
```
franky47: **suggestion**: could we show a check icon inline in the menu to indicate the copy?

For a couple of reasons:
- It keeps the behaviour colocated with the control
- It avoids pulling in sonner as a dependency

The downside is we have to reserve a place for the checkmark to be displayed (could be overlaid on the left icon?) and it requires keeping the popup open. Maybe too many trade-offs.

What do you think?

packages/docs/src/components/brand-assets-menu.tsx
```diff
@@ -0,0 +1,68 @@
...
+  previewClassName: string
+  svg: string
+}
+
+export type DownloadableBrandAsset = {
+  filename: string
+  svg: string
+}
+
+const assetsDir = path.resolve(process.cwd(), '../res')
```
franky47, line 25: **note**: I'll have to make sure this works after ISR, Vercel is/was having issues with relative paths in the ISR lambda:

https://francoisbest.com/posts/2023/reading-files-on-vercel-during-nextjs-isr

franky47 (approved):

packages/docs/src/components/brand-assets-menu.client.tsx
TheOrcDev: Yeah we could put a ghost button on the right side with copy icon that transforms to check to make it more obvious that it is clickable.

Let me see how it looks

TheOrcDev: Added button!

packages/docs/src/components/brand-assets-menu.tsx
TheOrcDev: Roger that!

packages/docs/src/components/brand-assets-menu.client.tsx
franky47: Perfect, we can remove the toast & sonner then.

franky47 (changes requested):

franky47 (approved):

PR #1417, opened 2026-06-01
chore: npm staged publishing

franky47:
Start phasing out Semantic Release for a two-phased approach that treats npm staged publishing as a first-class primitive (stage & open a draft GitHub release with generated notes).


franky47:
**note**: The handlebars critical CVE is pulled in via Semantic Release, which is what we're trying to phase out with this PR (part 1 of N).

PR #1442, opened 2026-06-18
chore: allow finalize to comment on related discussions

franky47:
In some cases I was bitten by semantic-release's inability to interpret closing references for discussions, not to close the discussion itself, but at least to be able to post a notification there when a release was made.

Don't mind the self-review thing, I haven't gone insane, I'm playing with loops to let an agent address my self-review as I post it.


.github/workflows/release-finalize.yml
```diff
@@ -28,11 +28,13 @@ jobs:
     # Least privilege: contents:read is checkout's documented minimum (the
     # `fetch-depth: 0` clone needs it; only public-repo anonymous read masks
     # its absence). Never contents:write, never an id-token — finalize only
-    # comments/labels. The npm publish authorization lives in release-draft.yml.
+    # comments/labels (issues & PRs via REST, discussions via GraphQL). The npm
+    # publish authorization lives in release-draft.yml.
```
franky47: **note**: we can drop the "the publish authorization..." sentence, it's irrelevant comment coupling.

packages/scripts/lib/commit-graph.ts
```diff
@@ -105,6 +105,33 @@ export function extractPRNumber(subject: string): number | null {
...
 }
 
+// --- Pure core: closing-keyword references --------------------------------
+
+// GitHub's closing keywords (case-insensitive), the ones that auto-link an
+// issue when written in a PR body: close/closes/closed, fix/fixes/fixed,
+// resolve/resolves/resolved.
+//   https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue
+const CLOSING_REFERENCE =
+  /\b(?:close[sd]?|fix(?:e[sd])?|resolve[sd]?)\b[:\s]+#(\d+)/gi
```
franky47: **note**: since the comment posted reads "This discussion is addressed...", we should also accept `Addresses #123` (and variants) as a reference. It's not a "closing" reference per se, but since we don't actually close discussions the term may be too narrow. TARGET_REFERENCE would follow our glossary better.

packages/scripts/lib/commit-graph.ts
```diff
@@ -370,6 +420,81 @@ function fetchClosingIssues(
...
+// error with a null node — *expected*, and dropped. This is the mirror image of
+// `fetchPRNodes`, which fails on any error: here `NOT_FOUND` is the normal "this
+// number isn't a discussion" signal, not a lost target.
+//
+// Only `NOT_FOUND` is benign. Every other error — including `FORBIDDEN` — is
+// thrown. Within a repo whose discussions you can read at all there is no
+// per-discussion visibility ACL, so a `FORBIDDEN` here is not a per-candidate "skip
+// this one" signal but a token/repository scope failure that nulls *every* alias —
+// exactly the case where silently returning `[]` would drop the whole feature on a
+// green run. Failing loud keeps a lost discussion from passing for "none".
```
franky47: don't comment on implementation details. reduce comments to strict explanations of why we do things, not how (we can read the code which is the SSOT for that).

packages/scripts/lib/commit-graph.ts
```diff
@@ -529,11 +675,13 @@ export async function discoverChanges(args: {
...
-// appear here. `fetchChangeDetails` is never touched. The PRs come from the
-// fetched (surviving) nodes, so a `(#N)` that resolves to nothing is dropped from
-// the targets exactly as it is from the notes.
+// comment targets — the release's PRs (by number), their deduplicated closing
+// issues, and the discussions resolved from the PR bodies' closing keywords.
+// Direct-commit changes have no PR to read a body from, so they contribute no
+// targets. `fetchChangeDetails` is never touched. The PRs come from the fetched
+// (surviving) nodes, so a `(#N)` that resolves to nothing is dropped from the
+// targets exactly as it is from the notes; likewise a discussion candidate that
+// no longer resolves is dropped by the probe.
```
franky47: "`fetchChangeDetails` is never touched" is unnecessary overcommenting.

PR #1446, opened 2026-06-22
ref: scope nullable useSearchParams defence to Next.js specific adapter

franky47:
There was some leftover null defence in useQueryStates from the time we called useSearchParams directly (pre-2.0.0), which can be moved into the app router adapter, so we enforce the searchParams adapter property non-nullable as a contract.

### DIFF
--- a/packages/nuqs/src/adapters/next/impl.app.browser.test.tsx
+++ b/packages/nuqs/src/adapters/next/impl.app.browser.test.tsx
@@ -1,0 +1,23 @@
+import { describe, expect, it, vi } from 'vitest'
+import { renderHook } from 'vitest-browser-react'
+import { useNuqsNextAppRouterAdapter } from './impl.app'
+
+// `useSearchParams()` returns `null` when rendered outside a SearchParamsContext
+// provider (e.g. `app/global-error` before Next 15.2, or isolated mounts/tests).
+// nuqs builds without a `pages/` dir, so it only sees the non-null app-router
+// overload — this null path can't surface from types, only by forcing it here.
+// The adapter must still expose a non-null `searchParams` (AdapterInterface).
+vi.mock('next/navigation.js', () => ({
+  default: {},
+  useRouter: () => ({ replace: vi.fn() }),
+  usePathname: () => '/',
+  useSearchParams: () => null
+}))
+
+describe('Next App Router Adapter', () => {
+  it('coalesces a null `useSearchParams()` into a non-null URLSearchParams', async () => {
+    const { result } = await renderHook(() => useNuqsNextAppRouterAdapter())
+    expect(result.current.searchParams).toBeInstanceOf(URLSearchParams)
+    expect(result.current.searchParams.size).toBe(0)
+  })
+})
--- a/packages/nuqs/src/adapters/next/impl.app.ts
+++ b/packages/nuqs/src/adapters/next/impl.app.ts
@@ -109,7 +109,11 @@ export function NavigationSpy() {
 
 export function useNuqsNextAppRouterAdapter(): AdapterInterface {
   const router = useRouter()
-  const searchParams = useSearchParams()
+  // `useSearchParams` is typed `ReadonlyURLSearchParams | null` for consumer
+  // apps that have a `pages/` directory (Next injects that compat overload
+  // through next-env.d.ts). Only truly null when rendered outside a
+  // SearchParamsContext provider: tests/mocks, or `app/global-error` (pre-15.2).
+  const searchParams = useSearchParams() ?? new URLSearchParams()
   const [optimisticSearchParams, setOptimisticSearchParams] =
     useOptimistic<URLSearchParams>(searchParams)
   const updateUrl: UpdateUrlFunction = useCallback((search, options) => {
--- a/packages/nuqs/src/useQueryStates.ts
+++ b/packages/nuqs/src/useQueryStates.ts
@@ -125,10 +125,9 @@ export function useQueryStates<KeyMap extends UseQueryStatesKeysMap>(
   const queuedQueries = debounceController.useQueuedQueries(
     Object.values(resolvedUrlKeys)
   )
-  const [internalState, setInternalState] = useState<V>(() => {
-    const source = initialSearchParams ?? new URLSearchParams()
-    return parseMap(keyMap, urlKeys, source, queuedQueries).state
-  })
+  const [internalState, setInternalState] = useState<V>(
+    () => parseMap(keyMap, urlKeys, initialSearchParams, queuedQueries).state
+  )
 
   const stateRef = useRef(internalState)
 
@@ -138,7 +137,7 @@ export function useQueryStates<KeyMap extends UseQueryStatesKeysMap>(
   // (optimistic) updates which don't immediately alter the URL source.
   const searchParamsSyncKey =
     Object.values(resolvedUrlKeys)
-      .map(key => `${key}=${initialSearchParams?.getAll(key)}`)
+      .map(key => `${key}=${initialSearchParams.getAll(key)}`)
       .join('&') + JSON.stringify(queuedQueries)
   // Adopts the current URL value into the internal state when it has changed.
   // Used both during render (below) and from the effect backstop further down.
@@ -190,8 +189,8 @@ export function useQueryStates<KeyMap extends UseQueryStatesKeysMap>(
           return [
             urlKey,
             parser?.type === 'multi'
-              ? initialSearchParams?.getAll(urlKey)
-              : (initialSearchParams?.get(urlKey) ?? null)
+              ? initialSearchParams.getAll(urlKey)
+              : (initialSearchParams.get(urlKey) ?? null)
           ]
         })
       )
@@ -430,8 +429,8 @@ function parseMap<KeyMap extends UseQueryStatesKeysMap>(
     const query =
       queuedQuery === undefined
         ? ((parser.type === 'multi'
-            ? searchParams?.getAll(urlKey)
-            : searchParams?.get(urlKey)) ?? fallbackValue)
+            ? searchParams.getAll(urlKey)
+            : searchParams.get(urlKey)) ?? fallbackValue)
         : queuedQuery
     if (
       cachedQuery &&


franky47:
AI-generated "blog post" of our research on Next.js implementation (⚠️ slop ahead ⚠️)

# When does `useSearchParams()` actually return null?

Next.js types `useSearchParams()` as `ReadonlyURLSearchParams | null` in some
projects. nuqs leans on that: the Next app-router adapter coalesces the value so
the rest of the library never has to think about null.

```ts
// packages/nuqs/src/adapters/next/impl.app.ts
const searchParams = useSearchParams() ?? new URLSearchParams()
```

The old comment above that line claimed the null came from "calling it in a
`pages/` route before the router is ready". That turned out to be wrong. This is
the story of chasing the real answer through every GA release of Next from 13.4
to 16.2, and what it taught us about our own test suite.


## The hook never changed

Start at the source. Here is `useSearchParams` from the published build:

```js
// next/dist/client/components/navigation.js
function useSearchParams() {
  const searchParams = useContext(SearchParamsContext)
  const readonlySearchParams = useMemo(() => {
    if (!searchParams) {
      // When the router is not ready in pages, we won't have
      // the search params available.
      return null
    }
    return new ReadonlyURLSearchParams(searchParams)
  }, [searchParams])
  // ...
  return readonlySearchParams
}
```

The context it reads has a null default:

```js
// next/dist/shared/lib/hooks-client-context.shared-runtime.js
const SearchParamsContext = createContext(null)
```

So the rule is short: `useSearchParams()` returns null exactly when
`SearchParamsContext` has no provider value above it. Nothing else.

The interesting part is that this body is the same in every GA release from 13.4.0
to 16.2.9. We fetched all 180 of them from unpkg and fingerprinted the function.
Every difference between versions is cosmetic or additive: the bailout call
signature moved around in 14.0 and 14.1, the `ReadonlyURLSearchParams` import path
shifted, 16.0 added a `useDynamicSearchParams?.()` call for cacheComponents, 16.2
added a build-time validation proxy. None of them touch the `if (!searchParams)
return null` line. The null condition has been frozen since the app router shipped.

That reframes the question. We no longer ask "when does the hook return null". We
ask "when is `SearchParamsContext` empty".


## Both routers always fill the context

On a normally rendered route, the provider is always there with a real value.

The Pages Router feeds it through an adapter that returns an empty
`URLSearchParams`, never null, while the router is warming up:

```js
// next/dist/shared/lib/router/adapters.js
function adaptForSearchParams(router) {
  if (!router.isReady || !router.query) {
    return new URLSearchParams()   // empty, never null
  }
  return asPathToSearchParams(router.asPath)
}
```

That value is wired in on both the client entry and during SSR:

```js
// next/dist/server/render.js  (and dist/client/index.js on the client)
<SearchParamsContext.Provider value={adaptForSearchParams(router)}>
```

So the "router not ready in pages" case that the hook comment names, and that the
nuqs comment parroted, does produce a falsy-looking result. It is an empty set of
params, not null. The hook returns `new ReadonlyURLSearchParams('')`, and reading
`get('x')` gives back `null` per key, but the object itself is there.

The App Router fills the context from the current URL, so it is also always a real
object:

```js
// next/dist/client/components/app-router.js
const { searchParams } = useMemo(() => {
  const url = new URL(canonicalUrl, typeof window === 'undefined'
    ? 'http://n'
    : window.location.href)
  return { searchParams: url.searchParams /* ... */ }
}, [canonicalUrl])

// ...
<SearchParamsContext.Provider value={searchParams}>
```

Conclusion so far: on a normal page render, in either router, the hook returns a
`ReadonlyURLSearchParams`. It does not return null there. So where does the null
people remember actually come from?


## The one place the provider is missing

A React context is only null when you read it from outside its provider. In a Next
app there is one normal-usage component that renders outside the whole provider
stack: the global error boundary, `app/global-error.js`.

Global error replaces the root layout when the tree above it throws. In the early
app router, the boundary that renders your `global-error` component sat outside
`<Router>`, and `<Router>` is where all the context providers live:

```js
// next/dist/client/components/app-router.js  @ 15.1.12  (and back to 13.4.0)
function AppRouter({ globalErrorComponentAndStyles: [globalErrorComponent], ... }) {
  return (
    <ErrorBoundary errorComponent={globalErrorComponent}>
      <Router /* providers are inside here */ />
    </ErrorBoundary>
  )
}
```

When that boundary trips, your `global-error` component renders as a sibling of
`<Router>`, above every provider. A client component in there that calls
`useSearchParams()` reads the `createContext(null)` default and gets null. That is
the real, normal-usage null.


## The fix landed in 15.2.0

This changed. We bisected the 15.x line version by version and pinned the exact
release.

```js
// next/dist/client/components/app-router.js  @ 15.2.0
function AppRouter({ globalErrorComponentAndStyles: [globalErrorComponent], ... }) {
  return (
    <ErrorBoundary errorComponent={DefaultGlobalError /* framework builtin */}>
      <Router globalError={[globalErrorComponent]} />
    </ErrorBoundary>
  )
}

// inside Router's render, the user boundary is now nested under the providers:
content = <ErrorBoundary errorComponent={globalError[0]}>{content}</ErrorBoundary>
return (
  <SearchParamsContext.Provider value={searchParams}>
    {/* other providers */}
    {content}
  </SearchParamsContext.Provider>
)
```

The boundary was split in two. A framework default error component now sits at the
very top as a last resort for when the router itself fails. Your own `global-error`
component moved down, nested inside `SearchParamsContext.Provider`. From 15.2.0 on,
`useSearchParams()` inside `app/global-error.js` returns a real object.

So the timeline is one line long:

- 13.4.0 to 15.1.12: `useSearchParams()` is null inside `app/global-error`
- 15.2.0 and later: it is non-null

Everything else stayed put. The pages "not ready" path, the custom error pages,
partial prerendering, cacheComponents: none of them return null. The newer ones
throw a client-side-rendering bailout or suspend on a pending promise instead, but
that is a different mechanism, and it surfaces as a build error or a missing
Suspense boundary, not as a null you read.


## The type is a separate story

There is a second reason people associate `useSearchParams` with null, and it has
nothing to do with runtime. It is the type.

```ts
// next/dist/.../navigation-types/compat/navigation.d.ts  (identical 13.4 to 16.2)
declare module 'next/navigation' {
  /**
   * If used from `pages/`, the hook may return `null` when the router is not ready.
   */
  export function useSearchParams(): ReadonlyURLSearchParams | null
}
```

This overload widens the return type to include null. It only applies when your
project pulls it in through `next-env.d.ts`:

```ts
// next-env.d.ts, emitted only for projects that have BOTH app/ and pages/
/// <reference types="next/navigation-types/compat/navigation" />
```

The condition that writes that reference is `hasAppDir && hasPagesDir`. So a pure
app-router project sees the non-null type. A hybrid project that still has a
`pages/` directory sees `| null`. The runtime is the same in both. The type is the
library being conservative about the pages-router not-ready window, even though the
adapter actually returns an empty object there.

That split is why a library like nuqs sees null at the type level. nuqs builds
without a `pages/` directory, so its own compile sees the non-null overload. A
consumer with a hybrid setup sees `| null`, and the adapter has to satisfy that
contract for them.


## What this means for the adapter

Put the two findings together and the guard earns its place:

```ts
const searchParams = useSearchParams() ?? new URLSearchParams()
```

It is doing real work in two situations, and only two:

1. A nuqs hook rendered with no Next provider above it. In practice that means
   isolated unit and component tests, or any mount outside the router tree.
2. A nuqs hook rendered inside `app/global-error.js` on Next 13.4 through 15.1.
   Within nuqs's supported range that is 14.2 through 15.1, a real window.

On every standard server render, client render, partial prerender, and
cacheComponents pass, the provider is present and the guard is a no-op. It is not
papering over the normal render path. It keeps `AdapterInterface.searchParams`
honestly non-null so the rest of nuqs can drop its per-read defenses, and it
covers the two cases above.

The corrected comment says exactly that, in two lines:

```ts
// Only truly null when rendered outside a SearchParamsContext provider:
// tests/mocks, or `app/global-error` (pre-15.2).
```


## The blind spot we found by accident

The investigation started because we tried to delete the guard. The honest test of
whether a line matters is to remove it and watch something fail.

We removed the `?? new URLSearchParams()` and ran the whole suite. Types passed.
Unit passed. Browser passed. Nothing failed. That is a problem: a line that nothing
defends can be deleted by anyone, in good faith, and the suite will wave it through.

The reason nothing caught it is layered:

- Types could not catch it. nuqs has no `pages/` directory, so its own compile sees
  the non-null overload. The nullable type only exists in consumer projects.
- The end-to-end tests could not catch it. They run inside a real provider, where
  the value is never null.
- There was no unit test for the app-router adapter at all, by design. The repo
  tests adapters through end-to-end runs.

The fix is a single browser test that forces the one condition the guard exists
for. Mock the hook to return null, render the adapter, assert the output is still a
real object:

```tsx
// packages/nuqs/src/adapters/next/impl.app.browser.test.tsx
vi.mock('next/navigation.js', () => ({
  default: {},
  useRouter: () => ({ replace: vi.fn() }),
  usePathname: () => '/',
  useSearchParams: () => null
}))

it('coalesces a null useSearchParams() into a non-null URLSearchParams', async () => {
  const { result } = await renderHook(() => useNuqsNextAppRouterAdapter())
  expect(result.current.searchParams).toBeInstanceOf(URLSearchParams)
  expect(result.current.searchParams.size).toBe(0)
})
```

Remove the guard now and this test goes red with "expected null to be an instance
of URLSearchParams". The line is defended.


## How we checked

A note on method, because the conclusions are only as good as the evidence.

We pulled the version list from the npm registry, then fetched the relevant files
straight from unpkg for every GA release from 13.4.0 to 16.2.9. For the hook body
we fingerprinted all 180 versions and grouped them into the handful of distinct
forms. For the provider wiring and the global-error boundary we read the actual
compiled output at the version boundaries, then bisected the 15.x line release by
release to land on 15.2.0.

Reading the published `dist/` rather than the GitHub source matters here. It is the
exact code that runs in a user's app, with no build step in between to second-guess.
The comments survive minification in these files, and one of them, "When the router
is not ready in pages, we won't have the search params available", is the line that
sent everyone down the wrong path for years. It describes an intent. The adapter
right next to it quietly returns an empty object instead.

PR #1451, opened 2026-06-25
feat: add history to adapters `defaultOptions`

manojVivek:
Implements the feature in this discussion: https://github.com/47ng/nuqs/discussions/1448


franky47 (approved):
LGTM, thanks! I'll add a disclaimer on the docs to point out that setting `history: 'push'` globally is likely to lead to navigation history pollution, and this will be good to go in the next release.

PR #1458, opened 2026-06-28
fix!: preserve percent-encoded separators in parseAsArrayOf

chatman-media:
While poking at `parseAsArrayOf`, I noticed it loses data when an item already contains the encoded form of the separator. With the default comma, serialize escapes `,` to `%2C`, but it leaves any pre-existing `%2C` in the value untouched. So a string like `"a%2Cb"` serializes to `a%2Cb,c`, and parsing that back decodes the `%2C` and splits on it, giving `["a,b", "c"]`.

```ts
const p = parseAsArrayOf(parseAsString)
p.parse(p.serialize(["a%2Cb", "c"])) // -> ["a,b", "c"]  (should be ["a%2Cb", "c"])
```

Same thing bites a bare `%` followed by something that happens to look like the encoding. The root cause is that `%` is the escape introducer but was never itself escaped, so the encoding isn't reversible.

Fix is to escape `%` -> `%25` before escaping the separator on serialize, and undo it in the reverse order on parse. The common case is unchanged (`["a", ",", "b"]` still serializes to `a,%2C,b`), it just makes the round-trip injective. Added a couple of cases to the existing `parseAsArrayOf` test for items carrying `%2C` and `%`.

### DIFF
--- a/packages/nuqs/src/parsers.test.ts
+++ b/packages/nuqs/src/parsers.test.ts
@@ -309,6 +309,19 @@ describe('parsers', () => {
     expect(parser.serialize(['a%2Cb', 'c'])).toBe('a%252Cb,c')
   })
 
+  it('parseAsArrayOf — backward compat with pre-#1458 URLs containing %25', () => {
+    // The percent-escaping fix (#1458) changes the on-the-wire format:
+    // `parse` now unconditionally undoes a `%25`->`%` escape that older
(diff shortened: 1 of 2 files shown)


franky47:
Thanks, I'll park this in the v3 release line as it's a breaking change for existing URLs (I added a failing test to confirm).

PR #1474, opened 2026-07-06
perf: type-check with the TypeScript 7 RC native compiler

franky47:
Adopts the [TypeScript 7.0 RC](https://devblogs.microsoft.com/typescript/announcing-typescript-7-0-rc/) native (Go) compiler for every CLI type-check step, while keeping the JavaScript TypeScript API on the 6.0 bridge for tooling that still imports it (tsdown, vitest, knip, Next). The native compiler ships no stable programmatic API until 7.1, so a straight bump would break those tools. The catalog splits the two roles so they coexist with no binary clash: `@typescript/typescript6` exposes `tsc6` (JS API), and the RC exposes `tsc` (native), which the type-check scripts pick up unchanged.

Type-checking gets 4–8× faster. Controlled local benchmark, native `tsc` vs JS `tsc6`, cold runs:

| package | JS (`tsc6`) | native (`tsc`) | speedup |
|---|---|---|---|
| `scripts` (`--noEmit`) | 1.01 s | 0.21 s | 4.9× |
| `docs` (`tsc`) | 2.92 s | 0.68 s | 4.3× |
| `e2e/react-router/v5` (`-b`) | 0.66 s | 0.09 s | 7.5× |
| `e2e/react` (`-b`) | 0.72 s | 0.09 s | 8.1× |

TS 7 drops some legacy compiler options, so `baseUrl`, `downlevelIteration` and `alwaysStrict` were removed from the affected tsconfigs. All were inert here: `paths` were already relative (resolved against the tsconfig dir), and nuqs emits through tsdown rather than `tsc`.

Note on CI numbers: `next` runs are heavily Turbo-cached, so job-level wall-clock understates the win — the table above is the cleaner measurement of the compiler itself.
---

**CI verified green** — run [28768438268](https://github.com/47ng/nuqs/actions/runs/28768438268), all 37 checks pass.

Job-level wall-clock is not a clean comparison: `next` runs are Turbo-cached while the PR runs cold, so most jobs pay a one-time cache-miss (Playwright executes for real). The type-check-bound jobs are the readable signal — even cold they match or beat the cached `next` baseline:

| job | `next` (cached) | this PR |
|---|---|---|
| CI (scripts) — vitest + `tsc` | 27 s | 24 s |
| Linting — knip type analysis | 66 s | 56 s |

The compiler speedup itself is the local benchmark above (4–8×); in CI it is partly absorbed by Turbo caching and dwarfed by install + Playwright time.

### DIFF
--- a/knip.json
+++ b/knip.json
@@ -38,7 +38,7 @@
         "src/adapters/testing.ts"
       ],
       "project": ["src/**/*.ts"],
-      "ignoreDependencies": ["@vitejs/plugin-react"]
+      "ignoreDependencies": ["@vitejs/plugin-react", "typescript-7"]
     },
     "packages/docs": {
       "entry": [
--- a/packages/docs/package.json
+++ b/packages/docs/package.json
@@ -82,6 +82,7 @@
     "shadcn": "^3.8.2",
     "shiki": "^3.22.0",
     "typescript": "catalog:typescript",
+    "typescript-7": "catalog:typescript",
     "unified": "^11.0.5",
     "vitest": "catalog:vitest"
   },
--- a/packages/docs/tsconfig.json
+++ b/packages/docs/tsconfig.json
@@ -1,6 +1,5 @@
 {
   "compilerOptions": {
-    "baseUrl": ".",
     "target": "ESNext",
     "lib": ["dom", "dom.iterable", "esnext"],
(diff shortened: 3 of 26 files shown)


franky47:
Superseded by #1484.

PR #1475, opened 2026-07-07
fix(key-isolation): unwatched keys leaking through filterSearchParams

franky47:
URLSearchParams iterators are live: deleting the current entry while iterating shifts the index, so the entry right after every deleted key is skipped and survives the filter.

filterSearchParams hit this whenever it mutates the same object it iterates (copy: false), which is the case in the react adapter's getSnapshot and in the react-router popstate handler. Unwatched keys leaked into a hook's filtered snapshot, breaking key isolation between hooks and causing unrelated key changes to trigger re-renders.

Fix: snapshot the keys with Array.from before deleting, detaching iteration from the live collection.

### DIFF
--- a/packages/nuqs/src/adapters/lib/key-isolation.test.ts
+++ b/packages/nuqs/src/adapters/lib/key-isolation.test.ts
@@ -1,0 +1,47 @@
+import { describe, expect, it } from 'vitest'
+import { applyChange, filterSearchParams } from './key-isolation'
+
+describe('applyChange', () => {
+  it('returns the same instance when watched keys are unchanged', () => {
+    const oldValue = new URLSearchParams('keep=5&other=1')
+    const newValue = new URLSearchParams('keep=5&other=2')
+    const result = applyChange(newValue, ['keep'], false)(oldValue)
+    expect(result).toBe(oldValue)
+  })
+
+  it('returns the filtered new value when watched keys changed', () => {
+    const oldValue = new URLSearchParams('keep=5')
+    const newValue = new URLSearchParams('a=1&b=2&c=3&d=4&keep=6&e=6')
+    const result = applyChange(newValue, ['keep'], false)(oldValue)
+    expect(result.toString()).toBe('keep=6')
+  })
+})
+
+describe('filterSearchParams', () => {
+  it('removes unwatched keys without leaking entries past a deleted key', () => {
+    const search = new URLSearchParams('a=1&b=2&c=3&d=4&keep=5&e=6')
+    const filtered = filterSearchParams(search, ['keep'], false)
+    expect(filtered.toString()).toBe('keep=5')
+  })
+
+  it('leaves the original untouched when copying', () => {
+    const search = new URLSearchParams('a=1&b=2&c=3&d=4&keep=5&e=6')
+    const filtered = filterSearchParams(search, ['keep'], true)
+    expect(filtered.toString()).toBe('keep=5')
+    expect(search.toString()).toBe('a=1&b=2&c=3&d=4&keep=5&e=6')
+  })
+
+  it('returns the same instance untouched when no keys are watched', () => {
+    const search = new URLSearchParams('a=1&b=2')
+    const filtered = filterSearchParams(search, [], false)
+    expect(filtered).toBe(search)
+    expect(filtered.toString()).toBe('a=1&b=2')
+  })
+
+  it('keeps all values of a multi-value watched key and drops unwatched ones', () => {
+    const search = new URLSearchParams('tag=a&x=1&tag=b')
+    const filtered = filterSearchParams(search, ['tag'], false)
+    expect(filtered.getAll('tag')).toEqual(['a', 'b'])
+    expect(filtered.has('x')).toBe(false)
+  })
+})
--- a/packages/nuqs/src/adapters/lib/key-isolation.ts
+++ b/packages/nuqs/src/adapters/lib/key-isolation.ts
@@ -32,7 +32,7 @@ export function filterSearchParams(
     return search
   }
   const filtered = copy ? new URLSearchParams(search) : search
-  for (const key of search.keys()) {
+  for (const key of Array.from(search.keys())) {
     if (!keys.includes(key)) {
       filtered.delete(key)
     }


franky47:
<details>

# Plan 001: Fix `filterSearchParams` mutating a `URLSearchParams` while iterating it

> maintain the index.
>
> **Drift check (run first)**: `git diff --stat dad2045e..HEAD -- packages/nuqs/src/adapters/lib/key-isolation.ts packages/nuqs/src/adapters/react.ts packages/nuqs/src/adapters/lib/react-router.ts`

## Status

- **Priority**: P1
- **Effort**: S
- **Risk**: LOW
- **Depends on**: none
- **Category**: bug

## Why this matters

nuqs (a React URL-query-state library) isolates each hook to the URL keys it
watches by filtering a `URLSearchParams` down to those keys. The filter
function deletes non-watched keys from the **same** `URLSearchParams` it is
iterating (when its `copy` argument is `false`). `URLSearchParams` iterators
are live: deleting the current entry shifts the iteration index, so the entry
immediately after every deleted key is **skipped** and survives the filter.

Reproduced in Node (same behavior in browsers):

```js
const p = new URLSearchParams('a=1&b=2&c=3&d=4&keep=5&e=6')
for (const key of p.keys()) {
  if (key !== 'keep') p.delete(key)
}
p.toString() // 'b=2&d=4&keep=5' — b and d leaked; expected 'keep=5'
```

Consequences: unwatched keys leak into a hook's filtered snapshot, so (a) the
snapshot's cache key (`filteredSearch.toString()`) changes when an *unwatched*
key changes, returning a new reference from `useSyncExternalStore` and
re-rendering hooks that should have been isolated, and (b) consumers of
`useOptimisticSearchParams` can observe foreign keys.

## Current state

- `packages/nuqs/src/adapters/lib/key-isolation.ts` — the buggy filter and its
  wrapper `applyChange`. Whole file (42 lines):

```ts
// key-isolation.ts:26-41
export function filterSearchParams(
  search: URLSearchParams,
  keys: string[],
  copy: boolean
): URLSearchParams {
  if (keys.length === 0) {
    return search
  }
  const filtered = copy ? new URLSearchParams(search) : search
  for (const key of search.keys()) {   // <-- iterates `search`
    if (!keys.includes(key)) {
      filtered.delete(key)             // <-- mutates `search` when copy === false
    }
  }
  return filtered
}
```

Note the bug only bites when `copy === false` (iterated object === mutated
object). When `copy === true`, iteration runs over the original `search` while
deletions hit the copy, so results are correct today — but only accidentally;
the fix below makes both modes robust.

- Call sites with `copy = false` (the affected paths):
  - `packages/nuqs/src/adapters/react.ts:79-83` — inside the
    `useSyncExternalStore` getSnapshot:

```ts
const filteredSearch = filterSearchParams(
  new URLSearchParams(location.search),
  watchKeys,
  false
)
```

  - `packages/nuqs/src/adapters/lib/react-router.ts:115-119` (useState
    initializer) and `:125` (`applyChange(new URLSearchParams(location.search), watchKeys, false)`
    in the popstate handler).
- Call sites with `copy = true` (unaffected, do not change):
  `react-router.ts:111` and `:131`.
- `applyChange` (`key-isolation.ts:4-24`) calls
  `filterSearchParams(newValue, keys, copy)` after an equality bail-out; it
  inherits the fix automatically.

Repo conventions: Prettier with no semicolons, single quotes, 2-space indent,
`arrowParens: avoid`, no trailing commas (see the `prettier` block in the root
`package.json`). Unit tests for `lib/` helpers are colocated
(`src/lib/compare.test.ts` is the exemplar) and run under the vitest `unit`
project in Node. Bundle size is a hard constraint (`size-limit`: Client 6 kB) —
keep the fix bytes-minimal.

## Commands you will need

|---|---|---|
| Browser tests | `pnpm --filter nuqs exec playwright install chromium && pnpm --filter nuqs test:browser` | all pass |
| Build + size gate | `pnpm --filter nuqs build && pnpm --filter nuqs test:size` | exit 0, within limits |

## Scope

**In scope** (the only files you should modify/create):
- `packages/nuqs/src/adapters/lib/key-isolation.ts`
- `packages/nuqs/src/adapters/lib/key-isolation.test.ts` (create)

**Out of scope** (do NOT touch, even though they look related):
- `packages/nuqs/src/adapters/react.ts` and
  `packages/nuqs/src/adapters/lib/react-router.ts` — the call sites are
  correct once the helper is fixed; changing their `copy` flags alters
  allocation behavior for no benefit.
- `packages/nuqs/vitest.config.ts` — the coverage `exclude` for
  `src/adapters/**` is a separate decision; do not edit it here.
- Any e2e bench under `packages/e2e/` — the existing `key-isolation` e2e
  specs must keep passing unmodified.

## Git workflow

- Branch: `fix/key-isolation-filter-mutation` off `next`
- Conventional commit, e.g. `fix: unwatched keys leaking through filterSearchParams`
- Do NOT add any Co-Authored-By/Claude attribution. Do NOT push or open a PR
  unless the operator instructed it.

## Steps

Create `packages/nuqs/src/adapters/lib/key-isolation.test.ts`, modeled
structurally on `packages/nuqs/src/lib/compare.test.ts` (vitest `describe`/
`it`/`expect`, no React needed). Cover:

1. **The bug**: `filterSearchParams(new URLSearchParams('a=1&b=2&c=3&d=4&keep=5&e=6'), ['keep'], false)`
   → `.toString()` is exactly `'keep=5'`.
2. Same input with `copy: true` → result is `'keep=5'` AND the original
   `URLSearchParams` still has all 6 keys.
3. `keys: []` returns the same instance untouched.
4. Multi-value watched keys survive:
   `filterSearchParams(new URLSearchParams('tag=a&x=1&tag=b'), ['tag'], false)`
   → `.getAll('tag')` is `['a', 'b']` and `x` is gone.
5. `applyChange`: returns the old value (same reference) when watched keys are
   unchanged; returns the filtered new value (no unwatched keys, per case 1's
   shape) when they changed.

**Verify**: `pnpm --filter nuqs test:unit` → the new file runs; case 1 (and 5's
changed-branch assertion) FAIL; cases 2–4 pass.

### Step 2: Fix `filterSearchParams`

Snapshot the keys before deleting, e.g.:

```ts
const filtered = copy ? new URLSearchParams(search) : search
for (const key of Array.from(search.keys())) {
  if (!keys.includes(key)) {
    filtered.delete(key)
  }
}
```

(`Array.from` (or `[...]`) detaches iteration from the live collection.
`URLSearchParams.delete(key)` removes all entries for that key, so duplicate
key names in the snapshot are harmless re-deletes.)

**Verify**: `pnpm --filter nuqs test:unit` → all pass, including the new file.

**Verify**, in order:
1. `pnpm --filter nuqs test:browser` → all pass (the react adapter's
   key-isolation behavior is exercised here).
2. `pnpm --filter nuqs test:types` → all pass.
3. `pnpm --filter nuqs build && pnpm --filter nuqs test:size` → within limits.
4. `pnpm lint:prettier` → exit 0.

## Test plan

Covered by Step 1 (5 cases in `key-isolation.test.ts`). Pattern:
`src/lib/compare.test.ts`. Verification: `pnpm --filter nuqs test:unit` → all
pass, including 5+ new assertions.

## Done criteria

- [ ] `pnpm --filter nuqs test:unit` exits 0; `key-isolation.test.ts` exists and passes
- [ ] `pnpm --filter nuqs test:browser` exits 0
- [ ] `pnpm --filter nuqs test:types` exits 0

## STOP conditions

- `key-isolation.ts` no longer matches the excerpt above (drifted).
- Any existing browser test fails after the fix — that would mean something
  relies on the leaked-keys behavior; the fix strategy needs maintainer input.
- `test:size` fails — the fix should be ~bytes; a size failure means something
  else changed.
- Test case 1 *passes before the fix* — the bug is already fixed upstream;
  report and mark the plan stale.

## Maintenance notes

- The planned single-store redesign will reshape URL observation; this filter
  is used inside adapter snapshots, so keep its no-alias-mutation guarantee if
  it is moved.
- Reviewer: check that the fix keeps the zero-copy fast path (`keys.length === 0`)
  and doesn't switch `copy: false` callers to copies (allocation per
  getSnapshot call matters).
- Deferred (audit finding, separate concern): `keys.includes` inside the loop
  is O(M×K); a `Set` would help, but only matters bundled with the broader
  shared-snapshot perf work.

</details>

PR #1477, opened 2026-07-07
fix: support multi-value search params in createSearchParamsCache

franky47:
compareSearchParams compared string[] values with !==, so equal-content arrays from different sources are never equal. Next.js can hand distinct-but-equal searchParams objects to different consumers in the same request (e.g. generateMetadata and the page), so any second parse() call on a multi-value param (?tag=a&tag=b) threw a spurious NUQS-501 "different inputs" error.

Fixes it by reusing the existing compareQuery helper (already used elsewhere for string[] value comparison) for a deep, element-wise comparison instead of reference equality.

Note: this means `createSearchParamsCache` now supports deep-comparison of arrays of search params, which it didn't before.

### DIFF
--- a/packages/nuqs/src/cache.test.ts
+++ b/packages/nuqs/src/cache.test.ts
@@ -193,6 +193,11 @@ describe('cache', () => {
         compareSearchParams({ q: 'x', empty: undefined }, { q: 'x' })
       ).toBe(false)
     })
+    it('rejects an absent key vs a key with an undefined value', () => {
+      expect(
+        compareSearchParams({ q: 'x' }, { q: 'x', empty: undefined })
+      ).toBe(false)
+    })
     it('accepts both values being undefined', () => {
       expect(compareSearchParams({ q: undefined }, { q: undefined })).toBe(true)
     })
--- a/packages/nuqs/src/cache.test.ts
+++ b/packages/nuqs/src/cache.test.ts
@@ -163,10 +163,38 @@ describe('cache', () => {
       const array = ['a', 'b']
       expect(compareSearchParams({ x: array }, { x: array })).toBe(true)
     })
-    it('does not do deep comparison', () => {
+    it('does a deep comparison of array values', () => {
       expect(compareSearchParams({ x: ['a', 'b'] }, { x: ['a', 'b'] })).toBe(
+        true
+      )
+    })
+    it('rejects different array values', () => {
+      expect(
+        compareSearchParams({ tag: ['a', 'b'] }, { tag: ['a', 'c'] })
+      ).toBe(false)
+    })
+    it('rejects different array lengths', () => {
+      expect(compareSearchParams({ tag: ['a'] }, { tag: ['a', 'b'] })).toBe(
         false
       )
+    })
+    it('rejects an array compared to a string', () => {
+      expect(compareSearchParams({ tag: ['a'] }, { tag: 'a' })).toBe(false)
+    })
+    it('accepts equal strings', () => {
+      expect(compareSearchParams({ q: 'x' }, { q: 'x' })).toBe(true)
+    })
+    it('accepts the same object reference', () => {
+      const p = { q: 'x' }
+      expect(compareSearchParams(p, p)).toBe(true)
+    })
+    it('rejects a key with an undefined value vs an absent key', () => {
+      expect(
+        compareSearchParams({ q: 'x', empty: undefined }, { q: 'x' })
+      ).toBe(false)
+    })
+    it('accepts both values being undefined', () => {
+      expect(compareSearchParams({ q: undefined }, { q: undefined })).toBe(true)
     })
   })
 })
--- a/packages/nuqs/src/cache.ts
+++ b/packages/nuqs/src/cache.ts
@@ -1,5 +1,6 @@
 import * as React from 'react'
 import type { SearchParams, UrlKeys } from './defs'
+import { compareQuery } from './lib/compare'
 import { error } from './lib/errors'
 import { createLoader, type LoaderFunctionOptions } from './loader'
 import type { inferParserType, ParserMap } from './parsers'
@@ -144,7 +145,7 @@ export function compareSearchParams(a: SearchParams, b: SearchParams): boolean {
     return false
   }
   for (const key in a) {
-    if (a[key] !== b[key]) {
+    if (!compareQuery(a[key] ?? null, b[key] ?? null)) {
       return false
     }
   }


franky47:
<details>

# Plan 003: Fix `compareSearchParams` false negatives on array-valued params (spurious NUQS-501)

> maintain the index.
>
> **Drift check (run first)**: `git diff --stat dad2045e..HEAD -- packages/nuqs/src/cache.ts packages/nuqs/src/cache.test.ts`

## Status

- **Priority**: P1
- **Effort**: S
- **Risk**: LOW
- **Depends on**: none
- **Category**: bug

## Why this matters

`createSearchParamsCache` (nuqs's Next.js App Router server-side cache) allows
`parse()` to be called more than once per request — e.g. in `generateMetadata`
*and* the page — by comparing the incoming `searchParams` object against the
one already parsed. That comparison uses `!==` per key. `SearchParams` values
can be `string[]` (repeated URL keys like `?tag=a&tag=b`), and Next.js can
pass distinct-but-equal object instances to different consumers in the same
request. Two equal arrays are never `!==`-equal, so the compare returns
`false` and the cache **throws** `NUQS-501 "Different inputs in the same
request"` — a spurious server error on any page using multi-value search
params with more than one `parse()` call.

## Current state

- `packages/nuqs/src/cache.ts` — server-only module (exported via
  `nuqs/server`), contains both the throw site and the comparator.

The throw path (`cache.ts:79-91`):

```ts
const c = getCache()
if (Object.isFrozen(c.searchParams)) {
  // Parse has already been called...
  if (c[$input] && compareSearchParams(searchParams, c[$input])) {
    // ...but we're being called with the same contents again, [...]
    return all()
  }
  // Different inputs in the same request - fail
  throw new Error(error(501))
}
```

The buggy comparator (`cache.ts:139-152`):

```ts
export function compareSearchParams(a: SearchParams, b: SearchParams): boolean {
  if (a === b) {
    return true
  }
  if (Object.keys(a).length !== Object.keys(b).length) {
    return false
  }
  for (const key in a) {
    if (a[key] !== b[key]) {   // <-- string[] values compare by reference
      return false
    }
  }
  return true
}
```

- `SearchParams` type (`packages/nuqs/src/defs.ts:3`):
  `Record<string, string | string[] | undefined>`.
- The repo already has a value-comparison helper for exactly this value shape:
  `packages/nuqs/src/lib/compare.ts` —

```ts
export function compareQuery<T extends Query>(a: T | null, b: T | null): boolean {
  if (a === b) return true                 // covers identical refs & both-null
  if (a === null || b === null) return false
  if (typeof a === 'string' || typeof b === 'string') return false
  if (a.length !== b.length) return false
  return a.every((value, index) => value === b[index]!)
}
```

  Note its semantics: equal strings return `true` via the `a === b` fast path;
  *different* strings return `false` at the `typeof` check; string-vs-array
  returns `false`; arrays compare element-wise. `Query` is
  `string | string[]` — it does not accept `undefined`, hence the `?? null`
  normalization below.

Repo conventions: Prettier — no semicolons, single quotes, 2-space,
`arrowParens: avoid`, no trailing commas. `cache.test.ts` (unit project, Node)
is the colocated test file; `compareSearchParams` is already exported, so it
can be tested directly.

## Commands you will need

|---|---|---|
| Build + size gate | `pnpm --filter nuqs build && pnpm --filter nuqs test:size` | exit 0, within limits (Server budget 3.8 kB) |

## Scope

**In scope** (the only files you should modify):
- `packages/nuqs/src/cache.ts` (the `compareSearchParams` function only)
- `packages/nuqs/src/cache.test.ts` (add tests)

**Out of scope** (do NOT touch):
- `packages/nuqs/src/lib/compare.ts` — reuse it; do not modify it.
- The `parseSync` / freeze / 501-throw logic in `cache.ts` — the throw for
  *genuinely* different inputs is correct and must remain.
- `packages/nuqs/src/loader.ts` — parses values but does not compare inputs.

## Git workflow

- Branch: `fix/cache-compare-array-values` off `next`
- Conventional commit, e.g. `fix: spurious NUQS-501 for multi-value search params in createSearchParamsCache`
- No Co-Authored-By/Claude attribution. Do NOT push or open a PR unless the
  operator instructed it.

## Steps

In `packages/nuqs/src/cache.test.ts`, add a
`describe('compareSearchParams', ...)` block importing `compareSearchParams`
from `./cache` (match the file's existing import style). Cases:

1. Equal arrays, different instances:
   `compareSearchParams({ tag: ['a', 'b'] }, { tag: ['a', 'b'] })` → `true`
   (**fails today**).
2. Different arrays: `{ tag: ['a', 'b'] }` vs `{ tag: ['a', 'c'] }` → `false`.
3. Different array lengths: `{ tag: ['a'] }` vs `{ tag: ['a', 'b'] }` → `false`.
4. Array vs string: `{ tag: ['a'] }` vs `{ tag: 'a' }` → `false`.
5. Equal strings: `{ q: 'x' }` vs `{ q: 'x' }` → `true` (regression guard).
6. Same reference: `const p = { q: 'x' }` → `compareSearchParams(p, p)` → `true`.
7. Key present with `undefined` vs absent key:
   `{ q: 'x', empty: undefined }` vs `{ q: 'x' }` → `false` (key-count check;
   regression guard — this behavior must not change).
8. Both `undefined`: `{ q: undefined }` vs `{ q: undefined }` → `true`.

**Verify**: `pnpm --filter nuqs test:unit` → case 1 FAILS, the rest pass.

### Step 2: Fix the comparator

Replace the loop body with a `compareQuery`-based comparison, normalizing
`undefined` to `null`:

```ts
import { compareQuery } from './lib/compare'
// ...
for (const key in a) {
  if (!compareQuery(a[key] ?? null, b[key] ?? null)) {
    return false
  }
}
```

(`undefined ?? null` → `null` on both sides hits `compareQuery`'s `a === b`
fast path → `true`; one-sided `undefined` → `null` vs value → `false`.)

**Verify**: `pnpm --filter nuqs test:unit` → all pass, including all existing
`cache.test.ts` tests.

**Verify**: `pnpm --filter nuqs test:types` → pass;
`pnpm --filter nuqs build && pnpm --filter nuqs test:size` → within limits
(watch the Server 3.8 kB budget — the added import is a few bytes);
`pnpm lint:prettier` → exit 0.

## Test plan

Covered by Step 1 (8 cases in `cache.test.ts`). Pattern: the existing
describe blocks in the same file. Verification:
`pnpm --filter nuqs test:unit` → all pass.

## Done criteria

- [ ] `pnpm --filter nuqs test:unit` exits 0; `compareSearchParams` tests exist and pass
- [ ] `pnpm --filter nuqs test:types` exits 0
- [ ] `grep -n 'a\[key\] !== b\[key\]' packages/nuqs/src/cache.ts` returns no matches

## STOP conditions

- `compareSearchParams` in `cache.ts` doesn't match the excerpt (drifted).
- Importing `./lib/compare` into `cache.ts` creates a circular import or a
  type error (`Query` vs `SearchParams` value mismatch that `?? null` doesn't
  resolve).
- The size-limit Server budget fails.

## Maintenance notes

- If `SearchParams` ever gains richer value types (e.g. numbers from a future
  Next.js API), `compareQuery` must be extended first — it currently assumes
  `string | string[]`.
- Reviewer: confirm the key-count check still precedes the loop (it is what
  catches keys present in `b` but not `a`).

</details>

packages/nuqs/src/cache.test.ts
```diff
@@ -163,10 +163,38 @@ describe('cache', () => {
...
+      expect(compareSearchParams({ q: 'x' }, { q: 'x' })).toBe(true)
+    })
+    it('accepts the same object reference', () => {
+      const p = { q: 'x' }
+      expect(compareSearchParams(p, p)).toBe(true)
+    })
+    it('rejects a key with an undefined value vs an absent key', () => {
+      expect(
+        compareSearchParams({ q: 'x', empty: undefined }, { q: 'x' })
+      ).toBe(false)
```
franky47, line 194: todo: we should add the opposite test (first argument with missing key, second with undefined) for symmetry.

PR #1479, opened 2026-07-07
fix: preserve hash fragment in createSerializer bases

franky47:
URL bases silently dropped the #hash fragment, and string bases folded it into the last query param value (later percent-encoded), corrupting links built off anchored pages such as pagination or share links.

Threads the hash through createSerializer's base-splitting logic so it survives both string and URL bases, and is preserved when clearing all params via a global null.

### DIFF
--- a/packages/nuqs/src/serializer.test.ts
+++ b/packages/nuqs/src/serializer.test.ts
@@ -249,4 +249,37 @@ describe('serializer', () => {
       expect(result).toBe('?a=1&foo=bar&z=1')
     })
   })
+  describe('hash preservation', () => {
+    it('keeps the hash with a string base and existing search', () => {
+      const serialize = createSerializer(parsers)
+      const result = serialize('/path?a=1#section', {})
+      expect(result).toBe('/path?a=1#section')
+    })
+    it('keeps the hash with a string base when adding a value', () => {
+      const serialize = createSerializer(parsers)
+      const result = serialize('/path#section', { str: 'foo' })
+      expect(result).toBe('/path?str=foo#section')
+    })
+    it('keeps the hash with a URL base', () => {
+      const serialize = createSerializer(parsers)
+      const url = new URL('https://example.com/path#section')
+      const result = serialize(url, { str: 'foo' })
+      expect(result).toBe('https://example.com/path?str=foo#section')
+    })
+    it('keeps a hash containing a ? character intact', () => {
+      const serialize = createSerializer(parsers)
+      const result = serialize('/p#frag?x', {})
+      expect(result).toBe('/p#frag?x')
+    })
+    it('keeps the hash when clearing all params with a global null', () => {
+      const serialize = createSerializer(parsers)
+      const result = serialize('/path?str=foo#section', null)
+      expect(result).toBe('/path#section')
+    })
+    it('leaves paths without a hash unchanged', () => {
+      const serialize = createSerializer(parsers)
+      const result = serialize('/path', {})
+      expect(result).toBe('/path')
+    })
+  })
 })
--- a/packages/nuqs/src/serializer.ts
+++ b/packages/nuqs/src/serializer.ts
@@ -67,9 +67,9 @@ export function createSerializer<
     arg1BaseOrValues: BaseType | Values,
     arg2values: Values | null = {}
   ) {
-    let [base, search] = isBase<BaseType>(arg1BaseOrValues)
+    let [base, search, hash] = isBase<BaseType>(arg1BaseOrValues)
       ? splitBase(arg1BaseOrValues)
-      : ['', new URLSearchParams()]
+      : ['', new URLSearchParams(), '']
     const values = isBase(arg1BaseOrValues) ? arg2values : arg1BaseOrValues
     if (values === null) {
       for (const key in parsers) {
@@ -79,7 +79,7 @@ export function createSerializer<
       if (processUrlSearchParams) {
         search = processUrlSearchParams(search)
       }
-      return (base + renderQueryString(search)) as Return
+      return (base + renderQueryString(search) + hash) as Return
     }
     for (const key in parsers) {
       const parser = parsers[key]
@@ -106,7 +106,7 @@ export function createSerializer<
     if (processUrlSearchParams) {
       search = processUrlSearchParams(search)
     }
-    return base + renderQueryString(search)
+    return base + renderQueryString(search) + hash
   }
   return serialize
 }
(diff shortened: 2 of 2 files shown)


franky47:
<details>

# Plan 004: Preserve the URL hash fragment in `createSerializer`

> maintain the index.
>
> **Drift check (run first)**: `git diff --stat dad2045e..HEAD -- packages/nuqs/src/serializer.ts packages/nuqs/src/serializer.test.ts`

## Status

- **Priority**: P2
- **Effort**: S
- **Risk**: LOW
- **Depends on**: none
- **Category**: bug

## Why this matters

`createSerializer` builds URLs/query strings from typed values, optionally on
top of a base (`string | URLSearchParams | URL`). Hash fragments in the base
are mishandled in both non-trivial base types:

- **`URL` base**: the hash is silently dropped —
  `serialize(new URL('https://x.com/path#section'), { q: 'a' })` →
  `https://x.com/path?q=a` (no `#section`).
- **`string` base**: the hash is folded into the query string —
  `serialize('/path?a=1#section', {})` treats `1#section` as the value of `a`,
  and the eventual re-render percent-encodes the `#`, producing
  `/path?a=1%23section`. The fragment becomes part of a param value.

Anyone building links off an existing URL (pagination on an anchored page,
share links) loses or corrupts the fragment. The nuqs e2e suite even has a
`hash-preservation` spec for the *hooks* — the serializer contradicts that
behavior.

## Current state

- `packages/nuqs/src/serializer.ts` (135 lines) — the whole feature.

The two return points that must carry the hash (`serializer.ts:74-83` and
`:106-110`):

```ts
if (values === null) {
  for (const key in parsers) {
    const urlKey = urlKeys[key] ?? key
    search.delete(urlKey)
  }
  if (processUrlSearchParams) {
    search = processUrlSearchParams(search)
  }
  return (base + renderQueryString(search)) as Return
}
// ... main loop ...
if (processUrlSearchParams) {
  search = processUrlSearchParams(search)
}
return base + renderQueryString(search)
```

The base splitter (`serializer.ts:122-134`):

```ts
function splitBase<BaseType extends Base>(base: BaseType) {
  if (typeof base === 'string') {
    const [path = '', ...search] = base.split('?')
    return [path, new URLSearchParams(search.join('?'))] as const
  } else if (base instanceof URLSearchParams) {
    return ['', new URLSearchParams(base)] as const // Operate on a copy of URLSearchParams, as derived classes may restrict its allowed methods
  } else {
    return [
      base.origin + base.pathname,
      new URLSearchParams(base.searchParams)
    ] as const
  }
}
```

Call site (`serializer.ts:70-73`):

```ts
let [base, search] = isBase<BaseType>(arg1BaseOrValues)
  ? splitBase(arg1BaseOrValues)
  : ['', new URLSearchParams()]
```

- `renderQueryString(search)` (from `./lib/url-encoding`) returns `''` or a
  string starting with `?`. A URL's serialization order is
  `path` + `?query` + `#hash`, so the hash must be appended **after**
  `renderQueryString`.
- Note `URL.hash` and the manual split both yield the fragment *including*
  the leading `#` (or `''` when absent) — keep that convention so appending is
  unconditional string concatenation.
- Hash may legally contain `?` (e.g. `/p#frag?x`), so for string bases split
  off the hash **before** splitting on `?`.

Repo conventions: Prettier — no semicolons, single quotes, 2-space,
`arrowParens: avoid`, no trailing commas. Tests: `serializer.test.ts`
(colocated, unit project, Node) — follow its existing describe/it structure.
Bundle budgets: serializer ships in both Client (6 kB) and Server (3.8 kB)
bundles — keep the change to a handful of bytes.

## Commands you will need

|---|---|---|
| Build + size gate | `pnpm --filter nuqs build && pnpm --filter nuqs test:size` | exit 0, within limits |

## Scope

**In scope** (the only files you should modify):
- `packages/nuqs/src/serializer.ts`
- `packages/nuqs/src/serializer.test.ts` (add tests)

**Out of scope** (do NOT touch):
- `packages/nuqs/src/lib/url-encoding.ts` (`renderQueryString`) — it renders
  the query string only; hash handling belongs to the serializer's base logic.
- Hook-level hash preservation (adapters) — already works, covered by e2e.

## Git workflow

- Branch: `fix/serializer-hash-preservation` off `next`
- Conventional commit, e.g. `fix: preserve hash fragment in createSerializer bases`
- No Co-Authored-By/Claude attribution. Do NOT push or open a PR unless the
  operator instructed it.

## Steps

Add a `describe('hash preservation', ...)` block to `serializer.test.ts`
(reuse the parsers/serializer setup already in the file). Cases, asserting
exact output strings:

1. String base with hash and existing search:
   `serialize('/path?a=1#section', {})` → `/path?a=1#section`.
2. String base with hash, adding a value:
   `serialize('/path#section', { <someKey>: <someValue> })` →
   `/path?<key>=<value>#section`.
3. `URL` base with hash:
   `serialize(new URL('https://example.com/path#section'), { <someKey>: <someValue> })`
   → `https://example.com/path?<key>=<value>#section`.
4. Hash containing a `?`: `serialize('/p#frag?x', {})` → `/p#frag?x`.
5. `values === null` clears params but keeps the hash:
   `serialize('/path?<key>=<value>#section', null)` → `/path#section`.
6. No hash → unchanged behavior (regression guard): pick one existing-style
   case, e.g. `serialize('/path', {})` → `/path`.

**Verify**: `pnpm --filter nuqs test:unit` → cases 1-5 FAIL, case 6 passes.

### Step 2: Thread the hash through

Extend `splitBase` to return a third element `hash` (`''` when absent):

```ts
function splitBase<BaseType extends Base>(base: BaseType) {
  if (typeof base === 'string') {
    const [pathAndSearch = '', ...hashParts] = base.split('#')
    const hash = hashParts.length ? '#' + hashParts.join('#') : ''
    const [path = '', ...search] = pathAndSearch.split('?')
    return [path, new URLSearchParams(search.join('?')), hash] as const
  } else if (base instanceof URLSearchParams) {
    return ['', new URLSearchParams(base), ''] as const // Operate on a copy of URLSearchParams, as derived classes may restrict its allowed methods
  } else {
    return [
      base.origin + base.pathname,
      new URLSearchParams(base.searchParams),
      base.hash
    ] as const
  }
}
```

Update the destructuring at the call site (`:70-73`) to
`let [base, search, hash] = ...` with `['', new URLSearchParams(), '']` as
the non-base fallback, and append `hash` at **both** return points:
`base + renderQueryString(search) + hash`.

**Verify**: `pnpm --filter nuqs test:unit` → all pass, including all
pre-existing serializer tests.

**Verify**: `pnpm --filter nuqs test:types` → pass;
`pnpm --filter nuqs build && pnpm --filter nuqs test:size` → within limits;
`pnpm lint:prettier` → exit 0.

## Test plan

Covered by Step 1 (6 cases). Pattern: existing `serializer.test.ts` blocks.
Verification: `pnpm --filter nuqs test:unit` → all pass.

## Done criteria

- [ ] `pnpm --filter nuqs test:unit` exits 0; hash-preservation tests exist and pass
- [ ] `pnpm --filter nuqs test:types` exits 0 (the `as const` tuple change must not leak into public types)

## STOP conditions

- `splitBase` doesn't match the excerpt (drifted).
- Any pre-existing serializer test asserts hash-dropping as expected behavior
  — that means the current behavior was chosen deliberately; needs maintainer
  input.
- `test:size` fails.

## Maintenance notes

- If a future option adds hash *writing* (e.g. `serialize(..., { hash })`),
  this three-tuple is the place to extend.
- Reviewer: check case 4 (hash containing `?`) — it's the reason the string
  base splits `#` before `?`.

</details>

PR #1493, opened 2026-07-19
chore: bump fumadocs to 16.14, fumadocs-mdx to 15.3, shiki to 4.4

franky47:
Dependency bump: fumadocs-core/ui 16.0.4 to 16.14.5, fumadocs-mdx 13.0.2 to 15.3.0 (two majors), shiki + @shikijs/transformers 3 to 4.4.3 (aligned with the shiki fumadocs now ships).

The mdx 13→15 migration: collections import from the generated `collections/server` entry (`toFumadocsSource()` replaces `createMDXSource`), `lastModifiedTime: 'git'` becomes the `lastModified` plugin, and `defineDocs` is exported whole instead of destructured so the generator emits a docs collection with its meta attached. next.config was already ESM.

fumadocs-ui 16.11 came with behavioural changes the diff adapts to:

- It no longer ships a `container` utility override; the site layouts rely on it for centering, so globals.css now provides it (same values: centered, 1400px cap on 2xl).
- `sidebar.banner` is a ReactNode now, not a component — passing the function crashed RSC serialization at build time.
- Loose page content under DocsLayout needs a `DocsPage` slot: the playground demo pages rendered as a collapsed grid without it.
- `Markdown` moved from `fumadocs-core/content` to `fumadocs-core/content/md`.

The 16.11 notebook rewrite also restyled the layout: navbar 64→56px, article padding 24px on mobile / 56px on xl, a margin on the first sidebar group label plus a 2px row gap, and the code copy button at `top-3`. Inline highlighted code became `code.shiki` instead of `span.shiki > code`, so our chip padding tweak stopped matching and wrapped chips overlapped on small screens. tweaks.css and `docsPageClassName` restore the previous rhythm; measured offsets (navbar, H1, first sidebar link, TOC title, copy button) now equal prod on desktop and mobile.

16.12 to 16.14 added more, all adapted to:

- 16.13 ships its own `D` theme hotkey, on by default in `RootProvider`. It ran beside the configurable hotkey from #1537 and kept toggling after a rebind or disable, so it is off.
- 16.14 swaps Orama for ZBSearch under the same `createSearchAPI`; search results match prod on the queries tried.
- shiki 4.4 no longer trims a blank line before a closing fence; three fences in server-side.mdx had one.

Visual regression pass: Playwright screenshots of every sitemap route (37) on desktop and mobile, light and dark, compared against production nuqs.dev, reviewed by a swarm of agents and then by hand. Left as upstream behaviour: the new stepped TOC rail, the mobile TOC bar naming the first heading at scroll 0, the sidebar bottom fade. Two pre-existing artifacts confirmed identical on prod: the degraded v2.9.0 changelog entry (its DTO has unrecognized `labels` keys) and the lowercase compact-number suffixes on /stats (Vercel's ICU renders `115m`, local Node `115M` — runtime, not code).

Still deferred: lucide-react 1.x (drops brand icons used on the conf pages), @types/node (pinned to the repo's Node 24), express 5 (e2e infra).


packages/docs/src/app/playground/(demos)/pagination/pagination-controls.client.tsx
```diff
@@ -29,7 +29,7 @@ export function ClientPaginationControls({
         <PaginationItem>
           <PaginationPrevious
             disabled={page === 1}
-            onClick={() => setPage(p => Math.max(1, p - 1))}
+            onClick={() => setPage((p: number) => Math.max(1, p - 1))}
```
franky47: **question**: Why do we need this type annotation? Type-safety should be ensured by nuqs.

packages/docs/src/app/globals.css
```diff
@@ -56,7 +69,12 @@
 @layer base {
   html,
   body {
-    height: 100%;
+    /*
+      min-, not plain height: fumadocs-ui 16.11 positions the docs
+      sidebar and TOC with `sticky` inside a grid that must grow with
+      the page content; a viewport-capped body freezes them in place.
+    */
+    min-height: 100%;
```
franky47, line 77: **todo**: check the impact of this on multiple viewport heights, including with bottom bars like iOS Safari or the X/Twitter webview.

packages/docs/mdx-components.tsx
```diff
@@ -8,16 +8,6 @@ import defaultMdxComponents from 'fumadocs-ui/mdx'
...
 
-declare module 'mdx/types.js' {
-  // Augment the MDX types to make it understand React.
-  namespace JSX {
-    type Element = React.JSX.Element
-    type ElementClass = React.JSX.ElementClass
-    type ElementType = React.JSX.ElementType
-    type IntrinsicElements = React.JSX.IntrinsicElements
-  }
-}
```
franky47, line 19: **question**: Why is this removed? Does Fumadocs have a built-in way to type-check JSX in MDX now?

packages/docs/src/app/docs/[[...slug]]/page.tsx
```diff
@@ -40,10 +42,12 @@ export default async function Page(props: PageProps<'/docs/[[...slug]]'>) {
   const gated = gatedHeadingIds(await page.data.getText('raw'), version =>
     isPublished(version, published)
   )
-  const toc = page.data.toc.filter(item => !gated.has(item.url.replace(/^#/, '')))
+  const toc = page.data.toc.filter((item: TOCItemType) => !gated.has(item.url.replace(/^#/, '')))
```
Comment (franky47) [inline packages/docs/src/app/docs/[[...slug]]/page.tsx]: **question**: Why the type annotation here? page.data.toc should be typed, if not we need to figure out why.

packages/docs/src/app/playground/(demos)/pagination/pagination-controls.client.tsx
franky47: Removed in 1c5fe45e. The annotation was masking an implicit-any error that only ever appeared locally (never in CI) and no longer reproduces from a clean state — stale `.next` typegen output from an earlier build. nuqs setter types flow correctly without it.

packages/docs/src/app/docs/[[...slug]]/page.tsx
Comment (franky47) [reply packages/docs/src/app/docs/[[...slug]]/page.tsx]: Same root cause as the pagination annotation — removed in 1c5fe45e. `page.data.toc` is properly typed `TOCItemType[]` by the fumadocs-mdx runtime types; the local-only implicit-any was stale `.next` typegen state and no longer reproduces.

packages/docs/mdx-components.tsx
franky47: It is @types/mdx (bumped to 2.0.14 in the first deps PR), not fumadocs: its `types.d.ts` now derives MDX JSX types from the global JSX namespace directly (an `any extends JSX.ElementType ? never : JSX.ElementType` probe), so with React types loaded, JSX in MDX type-checks against React.JSX out of the box. The manual augmentation became both redundant and a compile error (duplicate identifiers Element/ElementClass/ElementType/IntrinsicElements) once fumadocs-mdx 15 resolved the newer @types/mdx.

packages/docs/src/app/globals.css
franky47: Tested against the built app at 553px (X webview-ish), 664px (iOS Safari with toolbars), 874px (iPhone 16 Pro full), and desktop 600/1400px: sidebar and TOC stay pinned (top 56px) after deep scroll, the TOC column height tracks the viewport via fumadocs' own `--fd-docs-height: 100dvh` var, and there is no horizontal overflow at any height. Mobile hides both rails by design (drawer + popover). Caveat: headless emulation cannot reproduce the live toolbar collapse/expand cycle — but the change only relaxes the body height ceiling (`height: 100%` → `min-height`); the sticky math itself is fumadocs' dvh-based vars, identical to their reference sites.

PR #1504, opened 2026-07-24
feat: add `required` parser option for loaders

### LINKED ISSUE #1404: Required option for query params
Similar to `strict`, it would be handy with a `required` option that throws if the parameter is missing

```typescript
const searchParamsConfig = {
  count: parseAsInteger.withOptions({ required: true }),
}

const loadSearchParams = createLoader(searchParamsConfig )

// Expected:
const { count, contributors } = loadSearchParams(request, { strict: true })
//      count: number
```

Cauatn:
## Summary
- Parsers can opt into `withOptions({ required: true })` so `createLoader`/`loadSearchParams` throws when a key resolves to `null` (missing from the URL, or failed to parse) and there is no `defaultValue` to fall back to. A `defaultValue` always satisfies the requirement, even when the param itself is absent from the URL.
- Adds test coverage for the new option, including interaction with `strict` mode.
- Documents the new option in the server-side usage docs, gated behind `<SinceVersion v="2.10.0">` until released.

Closes #1404

## Test plan
- [x] `vitest run --project unit src/loader.test.ts` passes (26/26)
- [x] Typecheck clean on `loader.ts`

### DIFF
--- a/packages/docs/content/docs/server-side.mdx
+++ b/packages/docs/content/docs/server-side.mdx
@@ -234,6 +234,33 @@ loadSearchParams('?count=banana', { strict: true })
 // [nuqs] Error while parsing query `banana` for key `count`
 ```
 
+<SinceVersion v="2.10.0">
+
+### Required params
+
+<FeatureSupportMatrix introducedInVersion='2.10.0' />
+
+You can mark individual parsers as `required{:ts}`, so that the loader throws
+if the key resolves to `null{:ts}`: either the param is missing from the URL,
+or it failed to parse, and there is no [default value](/docs/basic-usage#default-values)
+to fall back to. A `defaultValue{:ts}` always satisfies the requirement, even
+when the param itself is absent from the URL.
+
+```ts
+const loadSearchParams = createLoader({
+  count: parseAsInteger.withOptions({ required: true })
+})
+
+// Throws: `count` is missing from the URL and has no default value
+loadSearchParams('')
+// [nuqs] Missing required query param `count`
+
+// Works fine: `count` is present and parses successfully
+loadSearchParams('?count=1')
+```
+
+</SinceVersion>
+
 ## Cache
 
 <FeatureSupportMatrix introducedInVersion='1.13.0' support={{ supported: true, frameworks: ['Next.js (app router)']}} />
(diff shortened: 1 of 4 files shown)


franky47:
Hum this feels redundant with `strict`, what's your actual use-case for this?

Cauatn:
> Hum this feels redundant with `strict`, what's your actual use-case for this?

`strict` throws when a value is present but invalid (e.g. `?count=banana`). It doesn't do anything when the key is missing entirely from the URL, in that case the loader just silently falls back to null (or the default, if one exists).

`required` targets: "this key must exist in the URL at all, not just be well-formed if it happens to be there."

A concrete case that motivated this: a page like `/orders/detail?orderId=4821`, where the page has no meaning without `orderId`. If that param drops off the URL for any reason (user edits the URL manually, a stale link [happen this week in the project], a redirect that forgets to forward the query string), `strict` won't catch it — there's no "invalid value" to complain about, just an absent one. Today that `null` silently flows into a DB query:

```ts
const { orderId } = await loadOrderParams(searchParams)
const order = await db.order.findUnique({ where: { id: orderId } })
// orderId is null here, depending on the ORM/DB this either throws a
// generic, unrelated error, or in the worst case returns unexpected data
```

The resulting error (if any) points at order.total being undefined three lines away, not at the actual root cause. required turns that into a clear, early failure right at the loader:

```ts
orderId: parseAsInteger.withOptions({ required: true })
// → "[nuqs] Missing required search param `orderId`"
```

So the two options are complementary rather than overlapping: `strict` = "if present, must be valid", `required` = "must be present at all".  Happy to help if need anything

franky47:
> a page like `/orders/detail?orderId=4821`, where the page has no meaning without `orderId`

I'd wager your problem is a **URL design** one: that `orderId` should be part of the pathname if it's semantically required.

A good example of this is the YouTube watch page: they encode the video ID in the `v` search param `/watch?v=dQw4w9WgXcQ`, when it should really be part of the pathname: `/watch/dQw4w9WgXcQ` (which is the case for their link shortener: `youtu.be/dQw4w9WgXcQ`).

Having required and optional search params (e.g. for YouTube, `v` and `t` to start at a given point) makes it hard to maintain over time.

PR #1518, opened 2026-08-01
fix: reset the testing adapter's URL update queue once per mount

### LINKED ISSUE #1517: NuqsTestingAdapter: resetUrlUpdateQueueOnMount runs on every render and (with hasMemory) aborts concurrently queued updates
## Context

```
"nuqs": "2.9.4"
```

What framework are you using?

- ✅ Next.js (app router) in the app itself
- ✅ The bug is in `NuqsTestingAdapter`, hit under Vitest 4.1.8 + jsdom with React 19.2.7

## Description

`NuqsTestingAdapter`'s `resetUrlUpdateQueueOnMount` prop (default `true`) calls `resetQueues()` in the component body, so despite its name it runs on **every render** of the adapter, not on mount:

```ts
// src/adapters/testing.ts
export function NuqsTestingAdapter({ resetUrlUpdateQueueOnMount = true, ... }) {
  ...
  if (resetUrlUpdateQueueOnMount) resetQueues()
```

Separately, with `hasMemory`, every flushed URL update re-renders the adapter (`updateUrl` calls `setSearchParams`). Combined, any state write that is enqueued in the same React batch as a previous flush landing gets **aborted by the adapter's own re-render**: `resetQueues()` calls `globalThrottleQueue.abort()`, the queued update never reaches the URL, and the abort's `queuedQuerySync` emit re-syncs every hook from the URL, which still holds the previous value. The in-memory state silently reverts.

## Reproduction

Two rapid writes to the same key, where the second lands in the same batch as the first flush's adapter re-render:

```tsx
// <NuqsTestingAdapter hasMemory> ... useQueryState("foo", parseAsString)
act(() => setFoo("bar"))          // enqueued, flushes (rateLimitFactor 0), adapter re-render pending
await screen.findByText("bar")    // first write landed
act(() => setFoo(null))           // enqueued; the adapter's render body then aborts the queue
// state reverts to "bar" and never becomes null; no later event ever corrects it
```

In our real case the key holds a sheet/dialog stack: open then close through two dispatches, and the close is dropped, so the dialog re-opens from the stale URL and stays open forever. Failure rate was timing dependent (roughly 1 in 15 locally, much higher on slow CI runners); wrapping the dispatches in `act` makes it deterministic, which is how we traced it.

`localStorage.debug = "nuqs"` trace of the failing close:

```
[nuq+ _r_4_ `sheets`] setState: { sheets: [] }
[nuqs gtq] Enqueueing sheets=null {
[nuqs] Aborting queues
[nuqs gtq] Resetting queue {"sheets":null}
[nuq+ _r_4_ `sheets`] State changed: { sheets: [ { sheet: 'add-webhook', entity: 'new' } ] }
```

Note the aborted queue still holds the update it is throwing away, and the hooks then revert to the URL's stale value.

## Workaround

`<NuqsTestingAdapter resetUrlUpdateQueueOnMount={false}>` fixes it for us; per-flush hygiene still holds through `autoResetQueueOnUpdate`.

## Suggested fix

Make the prop do what its name says, e.g. guard with a ref so the reset runs once per mounted instance:

```ts
const didReset = useRef(false)
if (resetUrlUpdateQueueOnMount && !didReset.current) {
  didReset.current = true
  resetQueues()
}
```

Happy to provide a full runnable reproduction if useful.

inf1nite-lo0p:
Fixes #1517.

`NuqsTestingAdapter` calls `resetQueues()` in the component body, guarded only by the `resetUrlUpdateQueueOnMount` prop, so despite the prop's name it runs on every render. With `hasMemory`, every flushed URL update re-renders the adapter (`updateUrl` calls `setSearchParams`). Combined, an update enqueued while that re-render is pending gets aborted by the adapter's own render: it never reaches the URL, and the abort's re-sync reverts every subscribed hook to the URL's stale value.

Following the contributing guide's suggestion, the first commit is the failing test and the second is the fix:

1. `test:` a browser test that writes a value, then chains a second write on the first write's promise. The promise resolves at the end of the flush, so the second write is enqueued in a microtask before the adapter's pending re-render commits. On `next` this fails with `onUrlUpdate` called once instead of twice: the second write is aborted by the adapter's render body and the state reverts.
2. `fix:` a ref guard so the reset runs once per mounted instance. The guard stays in the component body on purpose: moving it to an effect would run it after children's mount-time writes and abort those instead.

Verified locally: the new test fails on `next` without the fix and passes with it; `pnpm --filter nuqs test:unit` 262/262 and `test:types` pass; Prettier clean.

Where this was found, for context: a dashboard whose sheet stack lives in a `?sheets=` param. A close dispatched right after an open's flush was silently dropped about 1 run in 15 locally and far more often on loaded CI runners, leaving the sheet permanently open in tests. Deterministic standalone reproduction with traces: https://github.com/inf1nite-lo0p/nuqs-1517-repro


packages/nuqs/src/adapters/testing.browser.test.tsx
```diff
@@ -0,0 +1,49 @@
+import React from 'react'
+import { describe, expect, it, vi } from 'vitest'
+import { render } from 'vitest-browser-react'
+import { parseAsString } from '../parsers'
+import { useQueryState } from '../useQueryState'
+import { withNuqsTestingAdapter, type OnUrlUpdateFunction } from './testing'
+
+const waitForSettle = () =>
+  new Promise<void>(resolve => {
+    setTimeout(resolve, 100)
```
franky47, line 10: **nitpick**: Why 100ms here?

packages/nuqs/src/adapters/testing.browser.test.tsx
```diff
@@ -0,0 +1,49 @@
...
+  })
+
+describe('adapters/testing: update queue survival (#1517)', () => {
+  it('keeps an update enqueued while a flush lands (hasMemory)', async () => {
+    const onUrlUpdate = vi.fn<OnUrlUpdateFunction>()
+    let value: string | null = null
+    let setValue: (update: string | null) => Promise<unknown> = () =>
```
franky47, line 17: **nitpick**: I'd rather avoid closing on component internals, a more isolated way would be to render a test button to call the setter sequence and act on it.

franky47 (approved):
LGTM for the implementation, the test works but could use a more robust implementation.

Feel free to try the pkg.pr.new preview build to see if it solves your initial use-case.

PR #1520, opened 2026-08-03
chore: next 16.3.0 with typescript 7

franky47:
Next 16.3.0 is out of preview and its toolchain now works with TypeScript 7. This lets the repo move to a single `typescript` catalog entry: drop the `@typescript/typescript6` JS API bridge, the `typescript-7` alias in every package, and the knip exception that came with them.

One exception: the Next.js e2e bench builds against older Next.js versions in CI, which type-check through the TypeScript JS API that TS 7 does not provide. That package stays on the 6.0 bridge, with `experimental.useTypeScriptCli: false` so Next 16.3.0 takes the JS API path too (its new default, the TS CLI checker, needs a `tsc` binary the bridge does not expose).

### DIFF
--- a/knip.json
+++ b/knip.json
@@ -36,7 +36,7 @@
         "src/adapters/testing.ts"
       ],
       "project": ["src/**/*.ts"],
-      "ignoreDependencies": ["@vitejs/plugin-react", "typescript-7"]
+      "ignoreDependencies": ["@vitejs/plugin-react"]
     },
     "packages/docs": {
       "entry": [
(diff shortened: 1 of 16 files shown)


franky47:
Independent follow-up review also found no blockers. One non-blocking observation: Next.js 14/15 warn that `experimental.useTypeScriptCli` is unknown, then ignore it and pass using the TS6 API bridge. The config comment already documents this behavior; I would keep the simple unconditional option rather than add version-detection machinery solely to suppress warning noise.

PR #1528, opened 2026-08-05
fix: preserve explicit empty native arrays

franky47:
## Summary

- Preserve an explicit empty native array through serialization and loading.
- Keep the existing empty-string behavior for item parsers that accept an empty value.
- Add regression coverage for the empty array URL form.

### DIFF
--- a/packages/nuqs/src/parsers.ts
+++ b/packages/nuqs/src/parsers.ts
@@ -518,7 +518,10 @@ export function parseAsNativeArrayOf<ItemType>(
       const parsed = query
         .map((item, index) => safeParse(itemParser.parse, item, `[${index}]`))
         .filter(value => value !== null && value !== undefined) as ItemType[]
-      return parsed.length === 0 ? null : parsed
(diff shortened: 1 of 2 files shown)


packages/nuqs/src/parsers.ts
```diff
@@ -518,7 +518,11 @@ export function parseAsNativeArrayOf<ItemType>(
       const parsed = query
         .map((item, index) => safeParse(itemParser.parse, item, `[${index}]`))
         .filter(value => value !== null && value !== undefined) as ItemType[]
-      return parsed.length === 0 ? null : parsed
+      if (parsed.length > 0) {
+        return parsed
+      }
+      // An explicit empty array serializes as one empty query value.
```
franky47: The wording is weird here: we're in a parser, comments should be about parsing, not serialising.

franky47: Reworded this from the parser perspective: “Parse a single empty query value as an explicit empty array.”

PR #1546, opened 2026-08-12
fix: reset React adapter queues on popstate

franky47:
## Summary

- keep the React adapter queue-reset listener mounted for the adapter lifetime
- cancel pending URL updates on Back/Forward even after the last query subscriber unmounts
- exercise the default React adapter through the shared queue-lifecycle E2E coverage without opt-in History API synchronization

## Why

The hook-owned `popstate` subscription disappears when the last query subscriber unmounts. A pending throttled or debounced update can then survive browser navigation and write a stale URL afterward. The provider-owned listener stays active until the adapter itself unmounts.

## Corrections

- Merged the latest `next` changes, including the shared queue-lifecycle regression coverage from #1547.
- Replaced the hook-owned queue reset with a provider-owned listener so it survives zero query subscribers.
- Reused the shared queue-lifecycle fixture and removed the bespoke React-only fixture and timing assumptions called out by Fable.
- Kept React's normal `popstate` store subscription focused on rerendering; queue cancellation is now independent of hook subscriptions.


packages/e2e/react/src/routes/popstate-queue-reset.tsx
```diff
@@ -0,0 +1,40 @@
...
+      </button>
+      <label>
+        Query
+        <input
+          value={query ?? ''}
+          onChange={event => queueDebouncedValue(event.target.value)}
+        />
+      </label>
+      <output aria-label="Query value">{query}</output>
+      <output aria-label="Update status">{updateStatus}</output>
```
franky47: use the Display component here

packages/e2e/react/src/routes/popstate-queue-reset.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+  async function pushCurrentValue() {
+    setUpdateStatus('pending')
+    await setQuery('current', { history: 'push' })
+    setUpdateStatus('settled')
+  }
+
+  async function queueDebouncedValue(value: string) {
+    setUpdateStatus('pending')
+    await setQuery(value || null, {
```
franky47: don't use `|| null`, use a default value of `''` for the query state.

packages/e2e/react/src/routes/popstate-queue-reset.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+  return (
+    <>
+      <button type="button" onClick={pushCurrentValue}>
+        Push current value
+      </button>
+      <label>
+        Query
+        <input
+          value={query ?? ''}
```
franky47: don't do `?? ''`, use the default value instead (makes it non-nullable).

packages/e2e/react/src/routes/popstate-queue-reset.tsx
franky47: Addressed in 53dfbcd9: the fixture now uses the shared `Display` component, with the Playwright selectors updated to its stable IDs.

packages/e2e/react/src/routes/popstate-queue-reset.tsx
franky47: Addressed in 53dfbcd9: the query state now has a default value of `''`, and the setter receives the string directly.

packages/e2e/react/src/routes/popstate-queue-reset.tsx
franky47: Addressed in 53dfbcd9: the query state default makes it non-nullable, so the input now uses `value={query}` directly.

Issue #418, opened 2023-11-30
useQueryState with push mode is not updated after browser back


gkielwasser:
## Context

What's your version of `next-usequerystate`?
1.13.0

Next version: 14.0.3

Are you using:

- ❌ The app router
- ✅ The pages router

## Description

I'm using useQueryState  with the history push mode. I would expect using browser back button would update query state. It seems back is not taken in account and therefore the state is not updated.

## Reproduction

[...]

PR #1551, opened 2026-08-17
feat: seed SSR search params via serverSearch on the react adapter

danworkman1:
## Context

Follow-up to the discussion in #1425 (Astro adapter), where @franky47 suggested:

> If it can be done just by seeding a useState inside the React SPA adapter, adding a `serverSearch` prop might actually make it compatible with other setups, like Inertia/Django/Rails with a thin React "island", which would be a better outcome than an Astro-specific adapter.

This PR does exactly that — implemented on the `useSyncExternalStore` server snapshot rather than a `useState`, which turns out slightly neater: React renders from `getServerSnapshot()` both when server-side rendering **and** during the hydration render, so server and client markup agree by construction, with no extra state to reconcile. The existing note in `react.ts` anticipated this:

```ts
// Note: we could expose a getServerSnapshot() function to allow server-side rendering
// to let consumers wire that to their backend router (eg: in Astro SSR, Fastify, Hono etc).
```

## The problem

When the React SPA adapter is server-rendered (Astro island, Inertia page, `renderToString` behind Fastify/Hono/Express), there is no `location` to read from, so `getServerSnapshot()` returns empty search params and every query state renders its parser default. On deep links, the client then reads the real URL and disagrees with the server markup → React 418 hydration errors, plus a visible flash of default content.

## The fix

`NuqsAdapter` (from `nuqs/adapters/react`) gains an optional `serverSearch?: string | URLSearchParams` prop that seeds the server snapshot, filtered per-instance to the watched keys like the client snapshot, and cached for referential stability:

```tsx
<NuqsAdapter serverSearch={Astro.url.search}>
  <SearchApp />
</NuqsAdapter>
```

Omitted, behaviour is unchanged (empty search params on the server).

## Included

- `serverSearch` prop on the React SPA adapter (`packages/nuqs/src/adapters/react.ts`)
- Browser tests (`react.browser.test.tsx`): SSR seeding from string / no-leading-`?` string / `URLSearchParams`, default behaviour preserved when omitted, and a hydration test asserting a deep link hydrates with zero `console.error` calls (it fails with 3 errors if the seeding is removed)
- Docs: a "Server-rendered islands (Astro, Inertia, etc.)" section under the React SPA adapter
- `react-dom` added to `packages/nuqs` devDependencies (the new test renders with `react-dom/server` + `hydrateRoot` directly; only the types were present before)

## Production validation

This approach is extracted from a custom adapter running in production on an Astro + Cloudflare Workers app (deep-linked flight search with multiple islands sharing URL state), where it eliminated the #418 hydration errors and empty-state flash. Happy to adjust naming or scope — and to follow up with an `e2e/astro` test bench if that would be useful.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01GqkkYC3vJVgNtvK28pAsnw

### DIFF
--- a/packages/docs/content/docs/adapters.mdx
+++ b/packages/docs/content/docs/adapters.mdx
@@ -155,8 +155,8 @@ When the React tree is server-rendered by a framework nuqs has no adapter for
 — an Astro island, an Inertia page component, a server rendering the tree with
 `renderToString{:ts}` (Fastify, Hono, Express) — there is no `location{:ts}`
 to read from on the server, so query states render their default values.
-On deep links, this causes hydration errors and a flash of default content
-before the client corrects it.
+On deep links, this causes a flash of default content before the client
+corrects it once hydrated.
 
 Passing the request's search string to the adapter fixes this, by seeding
 the initial query states with it on the server, and letting React hydrate
--- a/packages/docs/content/docs/adapters.mdx
+++ b/packages/docs/content/docs/adapters.mdx
@@ -144,6 +144,8 @@ createRoot(document.getElementById('root')!).render(
 This may be useful for servers not written in JavaScript, like Django (Python),
 Rails (Ruby), Laravel (PHP), Phoenix (Elixir) etc...
 
+<SinceVersion v="2.10.0">
+
 ### Server-rendered islands <HumanContent><br className='hidden in-[#nd-toc]:block'/></HumanContent>(Astro, Inertia, etc.) [#server-rendered-islands]
 
 <FeatureSupportMatrix introducedInVersion='2.10.0' support={{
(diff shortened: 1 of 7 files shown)


franky47 (approved):
I wasn't able to reproduce the 418 hydration mismatch error, but the rest looks good. I merged the server cache with the one we already had for key isolation.

danworkman1:
Hi, @franky47, I ran the preview build against the app this was extracted from (Astro + Cloudflare Workers monorepo, two frontends sharing the search islands). Swapped our custom adapter for the stock `NuqsAdapter` with `serverSearch={Astro.url.search}` at the three island roots and deleted the custom adapter entirely.

Everything holds up. Typecheck/lint/unit all pass, and the full Playwright suite came back 178/181 — the 3 failures are known local load flakes that pass in isolation, and the failure set is identical to our baseline on the custom adapter. That includes our url-state roundtrip and invalid-URL hydration specs (junk params, malformed date ranges, unknown data and so on).

Deep links hydrate with zero console errors, and back/forward + cross-island updates all behave. Out of curiosity I also dropped `serverSearch` to see what breaks: no #418, which confirms the server-snapshot approach holds during the hydration render. You just get the default-content flash back, plus duplicated initial fetches (our transit lookup fired 3x instead of once). So the prop does exactly what it says on the tin.

Calling this fully validated on our side. Once it lands in a release we'll delete our adapter for good and swap the pkg.pr.new pin for the real version. Thanks for the quick turnaround on the preview build!

PR #1589, opened 2026-09-12
doc: add Sury parser page

DZakh:
Hi, Nuqs was a big inspiration for me when shaping the latest versions of Sury, and now that there's a stable release, I think Nuqs and Sury are a perfect match. I'd like to add a custom parser page to show how nicely the library fits Nuqs usecase. Let me know if you have any comments; I'm personally excited simply from looking at the code example.

Sury also implements Standard Schema, so parseAsJson takes a schema directly: covered alongside ArkType and Valibot.

### DIFF
--- a/packages/docs/content/docs/parsers/built-in.mdx
+++ b/packages/docs/content/docs/parsers/built-in.mdx
@@ -333,7 +333,7 @@ setJson({
 </HumanContent>
 
 Using other validation libraries is possible: `parseAsJson{:ts}` accepts
-any Standard Schema compatible input (eg: ArkType, Valibot),
+any Standard Schema compatible input (eg: ArkType, Valibot, Sury),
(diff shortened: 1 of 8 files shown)


franky47 (approved):
TIL about sury, the API design is really nice!

You might be interested in an RFC I'm drafting about StandardCodec, since sury seems to compose pretty well in one direction:

https://gist.github.com/franky47/f0d2cf0d39f799df6a5a31856207b840

The PR LGTM, we just have a cooldown period for new packages in pnpm so I'll merge this tomorrow when it clears for sury@11.0.0 which is quite recent.

DZakh:
Cool, I'll take a look 👍

Yeah, just shipped the version today. Decided to postpone it for a few days to add isEqual support specifically for the nuqs usecase  🙂

